PR #2, opened 2018-12-27
Fix server locked issues

psycho-ir:
I tested different conditions and driver behaves quite good, there were just 3 scenarios that `csi-driver` failed to do its job properly though.

1. When user wanna delete a `pvc` but the heztner volume is already attached to a server.
2. When user wanna attach multiple volumes to a pod. In this case csi-driver faced with `server locked` errors and fails as it returns `internal error`
3. When a pod schedules to execute on another node but some of the used volumes are already attached to another node. 

I fixed this issues in this PR and tested on a cluster.
LTM what your thoughts are about it. If you are OK with the implementation we can improve the tests.


thcyron:
First of all, thank you for your contribution!

Regarding your changes to `IdempotentService.Attach`: I’m not sure if we should detach the volume here. It may mess up the controller’s state. The [spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) says:

> This operation MUST be idempotent. If the volume corresponding to the volume_id has already been published at the node corresponding to the node_id, and is compatible with the specified volume_capability and readonly flag, the Plugin MUST reply 0 OK.
> 
> If the operation failed or the CO does not know if the operation has failed or not, it MAY choose to call ControllerPublishVolume again or choose to call ControllerUnpublishVolume.

To me, it reads like that: should the volume already be attached to a server, but it’s not the desired server, the CO will recover from that by detaching the volume via `ControllerUnpublishVolume`. The driver does not need to do that.

Regarding your changes to `IdempotentService.Delete`: I also don’t think we should detach here. After reading the spec for `DeleteVolume` again, I think we should return `FAILED_PRECONDITION` if the volume is still attached to a server.

If you want to continue contributing, I would kindly like to ask you to make smaller Pull Requests, ideally one change and one commit per PR. This makes reviewing much easier.

psycho-ir:
@thcyron sure thing, 

I just opened this PR quickly to discuss, I will open smaller PRs for each part. 
I'll add some comments to the code about the potential changes we can have in each PR, once we are in agreement I'll created the issue and open the PR, wdyt?

src/api/volume.go
```diff
@@ -100,6 +100,10 @@ func (s *VolumeService) Attach(ctx context.Context, volume *csi.Volume, server *
 		if hcloud.IsError(err, hcloud.ErrorCode("limit_exceeded_error")) {
 			return volumes.ErrAttachLimitReached
 		}
+		//TODO: update once hcloud upgraded to v1.12
```
psycho-ir: I think this part can be included in a PR about fixing the locked server errors CSI-driver receives from the hcloud.

src/api/volume.go
```diff
@@ -119,23 +123,16 @@ func (s *VolumeService) Detach(ctx context.Context, volume *csi.Volume, server *
...
 	if hcloudVolume.Server == nil {
 		return volumes.ErrNotAttached
 	}
-	if hcloudVolume.Server.ID != hcloudServer.ID {
-		return volumes.ErrAlreadyAttached
-	}
 
 	action, _, err := s.client.Volume.Detach(ctx, hcloudVolume)
 	if err != nil {
+		//TODO: update once hcloud upgraded to v1.12
```
psycho-ir: I think this part can be included in a PR about fixing the locked server errors CSI-driver receives from the hcloud.

src/driver/controller.go
```diff
@@ -169,6 +169,8 @@ func (s *ControllerService) ControllerPublishVolume(ctx context.Context, req *pr
 			code = codes.FailedPrecondition
 		case volumes.ErrAttachLimitReached:
 			code = codes.ResourceExhausted
+		case volumes.ErrLockedServer:
```
psycho-ir: I think this part can be included in a PR about fixing the locked server errors CSI-driver receives from the hcloud.

src/driver/controller.go
```diff
@@ -191,19 +193,15 @@ func (s *ControllerService) ControllerUnpublishVolume(ctx context.Context, req *
...
-
-	if err := s.volumeService.Detach(ctx, volume, server); err != nil {
+	if err := s.volumeService.Detach(ctx, volume); err != nil {
 		code := codes.Internal
 		switch err {
 		case volumes.ErrVolumeNotFound:
 			code = codes.NotFound
 		case volumes.ErrServerNotFound:
 			code = codes.NotFound
+		case volumes.ErrLockedServer:
```
psycho-ir: I think this part can be included in a PR about fixing the locked server errors CSI-driver receives from the hcloud.

src/volumes/service.go
```diff
@@ -14,6 +14,7 @@ var (
 	ErrAlreadyAttached     = errors.New("volume is already attached")
 	ErrNotAttached         = errors.New("volume is not attached")
 	ErrAttachLimitReached  = errors.New("max number of attachments per server reached")
+	ErrLockedServer        = errors.New("server is locked")
```
psycho-ir: I think this part can be included in a PR about fixing the locked server errors CSI-driver receives from the hcloud.

thcyron:
Handling `locked` errors from the Hetzner Cloud API and returning `ErrLockedServer` / `ABORTED` makes sense. 👍

I updated hcloud-go to v1.1.12 in master so you can use the error code constant.

PR #4, opened 2019-01-09
Handle server locked error

psycho-ir:
By returning codes.Aborted on `server locked` errors, CO will retry to publish/unpublish the volume.


thcyron:
Perfect! Thank you!

PR #11, opened 2019-02-05
Implement topology awareness

### LINKED ISSUE #10: Volume in different location than Pod
Hi, I am having an issue where the Volume seems to be created in one region and then the pod asking for it in a different region.
Here is the error:

```desc = failed to publish volume: server and volume must be in the same location```

I have a cluster with workers in different region, is this something the driver is not compatible with?

thcyron:
Fixes #10.

## Todo

* [x] Update storageclass manifest (WaitForFirstConsumer)
* [x] Update readme to enable feature gates for CSINoteInfo, etc.
* [x] Update readme to create CSINoteInfo CRD

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,34 @@ enabling you to use Volumes within Kubernetes. Please note that this driver **re
 
 ## Getting Started
 
-1. Create an API token in the [Hetzner Cloud Console](https://console.hetzner.cloud/).
+1. Make sure that the following feature gates are enabled in your cluster (kubelet and kube-apiserver):
 
-2. Create a secret containing the token:
+   ```
+   --feature-gates=CSINodeInfo=true,CSIDriverRegistry=true
+   ```
+
+   Also make sure that privileged pods are allowed in your cluster (kubelet and kube-apiserver):
+
+   ```
+   --allow-privileged=true
+   ```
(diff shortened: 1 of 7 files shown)


thcyron:
I pushed `hetznercloud/hcloud-csi-driver:1.1.0-rc.1` to Docker Hub with that change. Hope to be able to test that on a real cluster today before merging.

cc @mcadam

deacix:
How to use it? It's not working out of box. My volumes are deployed in the same location where the controller is running...

bascht:
Same here – running 1.14 and the version from master @ https://raw.githubusercontent.com/hetznercloud/csi-driver/master/deploy/kubernetes/hcloud-csi.yml and all volumes are deployed in the same location as the controller like in @deacix  case.

@thcyron is there a way to debug this or supply any helpful information?

thcyron:
To check if topology awareness is working:

1. Create a pod and schedule it on a node in a specific location
2. Create and attach a volume to that pod
3. Verify that the volume has been created in that location

Do this for every location you have nodes in. If pod creation failed, please provide the error messages.

bascht:
@thcyron Alright – I think I'm onto something. See #42. :smile:

Issue #60, opened 2019-09-25
Export Prometheus metrics


thcyron:
At least we should export some Go runtime metrics, probably also number of API calls per endpoint and error rates.

apricote:
I started working on this feature. You can find the progress in my fork on branch [`feature/prometheus`](https://github.com/apricote/csi-driver/tree/feature/prometheus).

To gather the gRPC metrics, I used the [go-grpc-prometheus](https://github.com/grpc-ecosystem/go-grpc-prometheus) library

I will create a PR once this is a bit more polished.

LKaemmerling:
Implemented with #67

PR #67, opened 2019-10-17
Add Prometheus support #60

apricote:
This PR will add support for exporting prometheus metrics.

It uses the [go-grpc-prometheus](https://github.com/grpc-ecosystem/go-grpc-prometheus) library to gather metrics from the `grpc.Server`.

Additionally it also exposes the default Go Runtime metrics (goroutines, memory, gc).

You can find an example of the exposed metrics in [this gist](https://gist.github.com/apricote/84ca8165255f022f0e5d46b47df4bd28).

## Known Issues

- The metrics endpoint for the DaemonSet is available on the public interface of the node. I am not sure why this is happening.


go.mod
```diff
@@ -4,21 +4,17 @@ go 1.13
 
 require (
 	github.com/container-storage-interface/spec v1.0.0
-	github.com/davecgh/go-spew v1.1.1 // indirect
```
apricote: I am not sure why there are so many changes in the `go.mod` and `go.sum` files. I executed `go mod tidy`, probably coming from that command.

cmd/driver/main.go
```diff
@@ -110,16 +112,35 @@ func main() {
 		os.Exit(1)
 	}
 
+	metricsEndpoint := os.Getenv("METRICS_ENDPOINT")
+	if metricsEndpoint == "" {
+		level.Error(logger).Log(
+			"msg", "you need to specify an endpoint for metrics via the METRICS_ENDPOINT env var",
+		)
+		os.Exit(2)
```
thcyron: In order not to introduce a breaking change, I would either:

1. make `METRICS_ENDPOINT` optional and not expose metrics if unset, or

2. provide a default if `METRICS_ENDPOINT` is unset.

I prefer option 2 and can’t think of a reason why defaulting to `:9092` would be a bad idea.

apricote: Only reason against using port `:9092` is the Prometheus ["Default port allocations"](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) list that recommends `> 9100` for exporters.

But i suspect that this will not be an issue because anyone can just remap the port through Docker should it actually cause a collision in their setup (non k8s).

thcyron: I just reused port 9092 from your code. Feel free to change it to a recommended one.

apricote: I reserved port `9189` for this, pushing updated code now.

apricote:
@thcyron Do you have an idea why the DaemonSet exposes the port on the public interface?

thcyron:
@apricote Probably due to: https://github.com/hetznercloud/csi-driver/pull/67/files#diff-e808103b0684391ddbc62202b863efafR175

Can we bind it to localhost?

LKaemmerling:
@apricote would it be possible to finish this PR within this or the next week :)? If this is not possible we would take over this PR.

LKaemmerling:
@apricote Thank you!

PR #72, opened 2019-11-21
Implement volume resizing and volume stats

LKaemmerling:


volumes/idempotency.go
```diff
@@ -150,3 +150,14 @@ func (s *IdempotentService) Detach(ctx context.Context, volume *csi.Volume, serv
 		return err
 	}
 }
+
+func (s *IdempotentService) Resize(ctx context.Context, volume *csi.Volume, size int) error {
+	switch err := s.volumeService.Resize(ctx, volume, size); err {
+	case ErrNotAttached:
+		return nil
```
thcyron: Why should errors be ignored when trying to resize a not attached volume?

PR #43, opened 2019-07-30
update deployment for k8s 1.14

costela:
- cluster-driver-registrar is no longer needed, instead being replaced
by manual declaration of the CSIDriver CRD.
- CSI sidecar images updated
- small README refactoring to account for version compatibilities

fixes #42

PR #82, opened 2019-12-03
remove deprecated csi-cluster-driver-registrar (again)

costela:
see: https://github.com/kubernetes-csi/cluster-driver-registrar
and:
https://kubernetes-csi.github.io/docs/csi-driver-object.html#what-creates-the-csidriver-object

This was added in #43 but reverted without explanation on 1a9edfaea5 (by
accident? What did I miss?)


LKaemmerling:
Thank you!

PR #120, opened 2020-06-05
added btrfs support

bhelm:
enables the use of btrfs as FSType


bhelm:
@LKaemmerling would you merge it?

PR #135, opened 2020-07-24
Added BTRFS Support

bhelm:
here again the PR for BTRFS support. Let me know if there is a reason not to merge it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,6 @@ ADD . /csi/
 RUN CGO_ENABLED=0 go build -o driver.bin hetzner.cloud/csi/cmd/driver
 
 FROM alpine:3.10
-RUN apk add --no-cache ca-certificates e2fsprogs xfsprogs blkid xfsprogs-extra e2fsprogs-extra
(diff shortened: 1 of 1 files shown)


LKaemmerling:
Hey, sorry for the late reply. btrfs is not supported by k8s for resizing volumes: See https://github.com/kubernetes/kubernetes/blob/master/pkg/util/resizefs/resizefs_linux.go#L60, so i don't see a big benefit from adding a filesystem type that would not be fully supported:
```
MountVolume.NodeExpandVolume failed : Expander.NodeExpand failed to expand the volume : rpc error: code = Internal desc = failed to resize volume: ResizeFS.Resize - resize of format btrfs is not supported for device /dev/disk/by-id/scsi-0HC_Volume_6393453 mounted at /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-59c793cd-06eb-4a6d-8cae-c19781efe9fa/globalmount
```

bhelm:
Ok, i see the point. I may do a PR at k8s for resizing support first and report back then.

PR #144, opened 2020-09-03
Update hcloud-csi-master.yml

ArtemLavrentii:
Added KUBE_NODE_NAME to controller. Now it should also get its instance id from node name.

Fixes a case when metadata service isn't available at all


LKaemmerling:
Thank you!

Issue #139, opened 2020-08-07
Still connecting to unix:///...


eBeyond:
The hcloud-csi nodes and the controller are in Crashloops after istalling hcloud-csi as described.

The hcloud-csi-driver reports:
`level=debug ts=2020-08-07T05:42:01.694472148Z msg="getting instance id from metadata service"`

All other pods of the hcloud-csi-controller reporting:
`Still connecting to unix:///var/lib/csi/sockets/pluginproxy/csi.sock`
The pods of the hcloud-csi-nodes are reporting (except the hcloud-csi-driver, which reports the same mentioned above):
`Still connecting to unix:///csi/csi.sock`

It's a k8s 1.18.6 running in nbg-dc3 on private networks only. All of them have no public network-interfaces but having a firewall with nat allowed for everything. ipv6 is disabled for all nodes.

eBeyond:
It seems that it's not possible to get an ID. It's calling http://169.254.169.254/2009-04-04/meta-data/instance-id (hardcoded in code).  But it's not reachable. Means: It cannot be reached from the k8s-nodes within the private network. It's only reachable from the nodes which have a public interface.

jekakm:
https://github.com/hetznercloud/csi-driver/issues/131
Take a look at this, if request are NATed - metadata server can't recognize what node making request and can't mount volume to appropriate server

eBeyond:
Okay.... That  seems to be a way. I'm not happy with it. Opening a door always make it accessible from both sides. But it's just a small one... :(

LKaemmerling:
@eBeyond try to use the current master deployment :) there is a small change in there which first uses the node name for getting the data.  https://github.com/hetznercloud/csi-driver/blob/master/deploy/kubernetes/hcloud-csi-master.yml

eBeyond:
I've tried the hcloud-csi-master, but it didn't succeed. Now I've tried it with the route and it work as expected. Thanks!

LKaemmerling:
Actually it should work, do the nodes in k8s have the same name as in the Cloud Console? Did you applied the new deployment?

JWDobken:
[...]

PR #147, opened 2020-09-08
Specify ndots dns config for hcloud-csi-driver

tumbl3w33d:
It doesn't hurt to be be explicit about it and avoids failure (e.g., with kubespray deployments).

Fixes one possible reason of #139

See https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html for more information.


LKaemmerling:
Hey @tumbl3w33d,

first of all, thank you for your PR. We don't see a big need to have this in the default deployment file, because the default file should be as basic as possible. Of course, feel free to change your deployment file and then deploy this.

- Lukas

PR #181, opened 2021-03-01
Microk8s support

vukomir:
changed hcloud-csi-controller from StatefulSet to DaemonSet so it can run on all nodes 

tested on microk8s 1.20 with multi-nodes


LKaemmerling:
Hey @vukomir,

sorry for the late answer (and the closing from the bot). 

We decided to not add additional k8s distributions for now, because we don't want to support every distribution in the future. We tried to make a good tradeoff between supporting what is requested and what we can easily support and the decision was made to support k8s (full-blown) and k3s officially. The CSI Driver might also work on microk8s, but we currently don't plan to support it officially.

PR #192, opened 2021-03-22
Fix image tag for v1.5.2

morremeyer:
With v1.5.2, the csi-driver starts to prefix the version with a `v`.

This PR fixes the deloyment yaml file not being conformant with this change.


LKaemmerling (approved):

LKaemmerling:
Thank you! I‘m sorry that this happened. I will look into why this happened tomorrow morning. I‘m sorry!

morremeyer:
@LKaemmerling Errors happen, no worries and no need to be sorry - on the contrary, thanks for being on top of things!

Issue #167, opened 2020-12-01
Volumes access mode information


filippoboscoUniTn:
Hello, I'm using hetzner for my kubernetes cluster and I'm currently studying volumes.

I was able to setup the hetzner's driver and dynamically provision persistent volumes via storage classes and persistent volume claims. I was wondering wether it was possible to have pods on different nodes access the same volume. Apparently for CSI volumes the access mode is driver specific ([doc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) and I could not find any information on hetzner's website. 

I'm not sure if it is appropriate to open an issue for such matters. What is the best place to obtain information and ask questions regarding hetzner's support for kubernetes?

LKaemmerling:
Hey @filippoboscoUniTn ,

our Volumes are ReadWriteOnce, so you can only access them by one pod at the same time. 

You can ask questions here (as Issues) or via the Support (Tickets) or within the Hetzner Discussion Board: https://forum.hetzner.com/ 

If you have any further questions, feel free to reopen the issue or to create a new issue.

PR #196, opened 2021-03-29
mention that volumes are ReadWriteOnce in the README.md

tpo:
Note the fact that the volumes are ReadWriteOnce. Which was an open question to which #167 did not know the answer (and neither did I nor my co-worker), so I think it'd be good to mention that characteristic of the Volume right there :-)!


pniederlag:
+1, should be noted somewhere

LKaemmerling:
Thank you!

PR #186, opened 2021-03-16
Add BTRFS support

fnkr:
Starting with v1.21, Kubernetes will fully support BTRFS volumes, including resize: https://github.com/kubernetes/kubernetes/pull/99361.

PR #200, opened 2021-04-21
Add btrfs support

LKaemmerling:
Replaces #186

Signed-off-by: Lukas Kämmerling <lukas.kaemmerling@hetzner-cloud.de>


fnkr:
I tried this today and it works great. Resizing works too.

fhofherr (approved):

felixscheinost:
For some reason this isn't working for me. How are you supposed to use this functionality?

I created a new `StorageClass` with `csi.storage.k8s.io/fstype: btrfs`. When I try to create a pod that uses that storage class I get an error: `MountVolume.SetUp failed for volume "pvc-eebbbe87-469b-4aba-a17c-9b080160051c" : rpc error: code = Internal desc = failed to pub │
│ lish volume: unsupported disk format btrfs`

Looking for `unsupported disk format` I found `FormatDisk` in `volumes/mount.go`, where `btrfs` isn't handled yet. Is this an error? Do the disks need to be preformatted? Do I need to set some other parameter?

PR #205, opened 2021-06-12
Separate server ID from location

NotTheEvilOne:
This PR adds support to execute the CSI plugin on a node not hosted on the Hetzner Cloud. It adds a new environment variable `HCLOUD_LOCATION_NAME` used for the controller service to work properly even if no server ID was determined. The node service is deactivated in this case.

A server ID (> -1) found in the environment variable or with the metadata service is always preferred before `HCLOUD_LOCATION_NAME` is considered.

This is useful in a Gardener deployment where the "csi-driver" is not necessarily deployed on a node in the Hetzner Cloud offering but still manages volumes for nodes deployed on the cloud.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,11 +76,17 @@ enabling you to use ReadWriteOnce Volumes within Kubernetes. Please note that th
 
 ## Integration with Root Servers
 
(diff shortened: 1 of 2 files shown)


NotTheEvilOne:
Please let me know if you see any issues @LKaemmerling or @fhofherr.

PR #215, opened 2021-08-02
Run e2e tests in parallel.

samcday:
This significantly sped up the e2e suite in CI, you can see an example [here](https://github.com/samcday/csi-driver/actions/runs/1089325239). Looks to be about a 2x speedup.


LKaemmerling:
Good catch! Thank you!

PR #221, opened 2021-08-19
CI improvements for speed and fork-friendliness.

samcday:
Cache Golang modules with actions/cache. This speeds up the unit tests
workflow about 3x (from ~1m30s to ~20s).

.dockerignore to ensure only code changes cause Docker context to change
and require a Golang rebuild.

Switch Docker builds to use official Docker images powered by BuildKit.
These builds run in parallel, and also cache layers between workflow
executions. The new builder, if hitting a fully primed cache, can finish
in 30 seconds (down from ~2-3m).

Images are tagged by scoping the Docker Hub organization to the repo
owner. Thus it's possible to fork the project and set DOCKER_USER +
DOCKER_PASSWORD in the fork environment and produce release artifacts
easily.

---

You can see some examples of the updated tag release CI here: https://github.com/samcday/csi-driver/actions/workflows/publish_on_tag.yml

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+.dockerignore
+Dockerfile
+.git/
+.github/
+.gitlab-ci.yml
+.idea/
+CHANGES.md
+LICENSE
+README.md
+e2etests/
+deploy/
+mock/
+script/
+**/*_test.go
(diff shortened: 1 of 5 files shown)


LKaemmerling:
Good catch! Thank you!

PR #222, opened 2021-08-19
Simplify CSI socket handling

samcday:
The various sidecars all default to looking for /run/csi/socket. By
changing the driver to respect that convention, we can eliminate a bunch
of unneeded configuration params in the deployment manifests.


LKaemmerling:
Hey @samcday,

thank you again for a great PR. I just rebased your MR and will now test it a bit locally to see if everything is fine :) I will come back in ~ 20 minutes ;)

samcday:
@LKaemmerling cool!

[Per my comment here](https://github.com/hetznercloud/csi-driver/pull/224#issuecomment-901926311) it'd be great if e2e tests ran on PRs like this, it's a lot less labour intensive than running them from my dev setup! :)

LKaemmerling:
@samcday thank you it works :) I will think about improving the test setup even more :)

PR #223, opened 2021-08-19
Kustomization support and manifest simplification

samcday:
kustomization.yaml makes it possible to directly include the repository
as a github.com URL, and also reference a particular tag (albeit only
from the next tagged release onwards). i.e

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
    - github.com/hetznercloud/csi-driver?ref=v0.6.0 # or ?ref=master
```

hcloud-csi-master.yml has been folded into hcloud-csi.yml. When a
release is tagged, the tag is amended with a hcloud-csi.yml that points
at the images for that specific version.


samcday:
If you don't like the "forcibly overwrite the newly-pushed tag with an update" approach, we could also create a release script that makes that change *before* the tag is pushed. Let me know what you think.

samcday:
Before I pushed 7145ece just now, this PR was missing some stuff. Sorry about that!

I think it's in better shape now.

.github/workflows/publish_on_tag.yml
```diff
@@ -10,6 +10,19 @@ jobs:
...
+        env:
+          RELEASE_VERSION: ${{ steps.release_version.outputs.value}}
+          OWNER: ${{ github.repository_owner }}
+        run: |
+          sed -i -e "s#: .*/hcloud-csi-driver:latest#: $OWNER/hcloud-csi-driver:$RELEASE_VERSION#" deploy/kubernetes/hcloud-csi.yml
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add .
+          git commit -m "Update image tags in hcloud-csi.yml to $RELEASE_VERSION"
+          git tag -d v$RELEASE_VERSION
```
LKaemmerling, line 23: iirc the `v` is to much here. The $RELEASE_VERSION is always the tag, and that is something like `v1.5.0`. So adding `v` here and below is not needed.

LKaemmerling (approved):

PR #234, opened 2021-08-26
e2e test workflow improvements

samcday:
The concurrency changes bring test time down to ~15m, [as you can see here](https://github.com/samcday/csi-driver/actions/runs/1168306054).

I've noticed that pushing the concurrency any higher seems to trigger rate limiting or some kind of problem on the hcloud platform side - I see lots of attach/detach API calls that fail with "unknown_error".

A concurrency of 3 per node seems to be fairly stable. Could maybe push this concurrency higher since we have 3 nodes but I've noticed that scheduler seems reluctant to place the e2e test pods on the control plane node, so better to assume there's only 2 nodes available with a 3 node setup.

Also tried to setup e2e test runs for PRs here, but reverted it because fork workflow runs don't have access to secrets. There are ways to work around this, [see this GitHub blog for example](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). What do you think about setting something like that up?

### DIFF
--- a/.github/workflows/test_e2e.yml
+++ b/.github/workflows/test_e2e.yml
@@ -1,9 +1,5 @@
 name: Run e2e tests
-on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+on: [push]
 concurrency: ci-${{ github.ref }}
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


LKaemmerling (approved):

LKaemmerling:
Thank you! We currently work on a safe solution to enable PRs with tests from customers.

PR #235, opened 2021-08-26
Remove unneeded privileges from CSI sidecars

samcday:
[Green e2e build here](https://github.com/samcday/csi-driver/actions/runs/1169866312)


LKaemmerling:
Thank you!

PR #237, opened 2021-08-27
Include more information in request logging

samcday:
Opted to filter out logging for Probe requests since the livenessprobe sidecar will freak out in informative ways if there's anything going wrong with that API call.


LKaemmerling:
Hey @samcday,

thank you!

PR #242, opened 2021-09-22
Update hcloud-csi.yml

sui77:
Adding same nodeAffinity for hcloud-csi-controller as hcloud-csi-node, since it doesn't seem to run well on root servers, too.

### DIFF
--- a/deploy/kubernetes/hcloud-csi.yml
+++ b/deploy/kubernetes/hcloud-csi.yml
@@ -110,6 +110,15 @@ spec:
       labels:
         app: hcloud-csi-controller
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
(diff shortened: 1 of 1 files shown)


fhofherr (approved):

fhofherr:
Hi,

sorry for the late reply and thank you for your contribution. Your PR looks good to me. However, before we merge: could you share which problems you ran into when the `hcloud-csi-controller` ran on a root server?

sui77:
IIRC the hcloud-csi-node pod from the daemonset didn't run on dedicated hosts at all and k8s retried to spin them up again and again. Don't remember the error message and it's not a real issue, I just wanted to remove the warnings from my k8s dashboard :)

It's not possible to mount cloud drives to dedicated hosts, right? Otherwise I'd rather figure out why it didn't run.

fhofherr:
Correct. It is currently not possible to mount cloud drives to dedicated hosts. Thanks again for your contribution.

PR #245, opened 2021-09-27
Add --csi-address to liveness-probe

### LINKED ISSUE #232: Upgrade Problems
Hello,

how can i upgrade the csi provider? i tried to apply the new 1.6.0 yaml, but than all pods are in a crash loop, because the healthchecks failed, maybe because of the new socket paths? I used RKE with Kubernetes Version v1.20.9

Thanks

vaisov:
Closes #232

### DIFF
--- a/deploy/kubernetes/hcloud-csi.yml
+++ b/deploy/kubernetes/hcloud-csi.yml
@@ -190,6 +190,8 @@ spec:
         - name: liveness-probe
           imagePullPolicy: Always
           image: k8s.gcr.io/sig-storage/livenessprobe:v2.3.0
+          args:
+            - --csi-address=/run/csi/socket         
           volumeMounts:
             - mountPath: /run/csi
               name: socket-dir
--- a/deploy/kubernetes/hcloud-csi.yml
+++ b/deploy/kubernetes/hcloud-csi.yml
@@ -191,7 +191,7 @@ spec:
           imagePullPolicy: Always
           image: k8s.gcr.io/sig-storage/livenessprobe:v2.3.0
           args:
-            - --csi-address=/run/csi/socket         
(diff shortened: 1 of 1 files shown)


LKaemmerling:
Hey @vaisov,

thank you for your MR. Could you please add the same to the `hcloud-csi-master.yml` ?

r-ising:
Hey @LKaemmerling  and @vaisov,
you don't need to add the csi-address, for upgrade reason its okay to set args to an empty array:

```
        - name: csi-attacher
          image: k8s.gcr.io/sig-storage/csi-attacher:v3.2.1
          args: []
```

```
        - name: csi-resizer
          image: k8s.gcr.io/sig-storage/csi-resizer:v1.2.0
          args: []
```

```
        - name: liveness-probe
          imagePullPolicy: Always
          image: k8s.gcr.io/sig-storage/livenessprobe:v2.3.0
          args: []
```

It's an temporary fix, you can remove it after upgrade, so i dont know, if you want it here..

samcday:
Thanks for your contribution @vaisov 

As @r-ising pointed out, setting csi-address is not necessary as the [various](https://github.com/kubernetes-csi/external-attacher/blob/master/cmd/csi-attacher/main.go#L55) CSI [external](https://github.com/kubernetes-csi/external-resizer/blob/master/cmd/csi-resizer/main.go#L55) sidecars [use](https://github.com/kubernetes-csi/livenessprobe/blob/master/cmd/livenessprobe/main.go#L44) `/run/csi/socket` by default.

The issue with the 1.6 upgrade was precisely what @r-ising stated: the complete absence of the `args` meant that `kubectl apply`ing the 1.6 manifests over the top of an existing <1.6 installation would preserve the original args. We'll be mindful and careful of this in the future! :)

PR #249, opened 2021-10-29
driver: fix panic when server is not found.

### LINKED ISSUE #248: panic: driver panics when `Server.GetByID` returns not found in main
The main routine can panic when attempting log that the target server has been fetched. I believe this occurs when the target server returns a `not_found` error; which is [translated by the client to a nil server and nil error response](https://github.com/hetznercloud/hcloud-go/blob/master/hcloud/server.go#L196-L201). The driver therefore also should also check for a nil server response.

Stack trace:
```
goroutine 1 [running]:
main.main()
	/csi/cmd/driver/main.go:125 +0x12c1
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x905701]
```

Nomad issue: https://github.com/hashicorp/nomad/issues/11405

jrasell:
closes #248

### DIFF
--- a/cmd/driver/main.go
+++ b/cmd/driver/main.go
@@ -122,6 +122,14 @@ func main() {
 		)
 		os.Exit(1)
 	}
+
+	// Cover potential cases where the server is not found. This results in a
+	// nil server object and nil error. If we do not do this, we will panic
+	// when trying to log the server.Name.
+	if server == nil {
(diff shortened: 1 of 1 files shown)


LKaemmerling (approved):
Thank you, good catch!

PR #270, opened 2022-01-30
[8/???] Remove Node Stage/Unstage capability

samcday:
Since b770ef5 the desired filesystem type is provided to the hcloud API.
As such the formatting no longer happens during the Stage call, making
it entirely unnecessary. Instead, the prepared volume is mounted to the
target location kubelet ultimately wants in the NodePublishVolume RPC
method.

Even if we for some reason wanted to do more fancy stuff on the node
side (I hope not, the whole point of the current patch series is to
reduce the node behaviour down to as minimal as possible), I think that
could just be done during the publish call.

Honestly, it's a bit of a head-scratcher why the stage/unstage
capability exists at all. I guess there's some use case for it, but it
doesn't appear our needs are complex enough to require it.

### DIFF
--- a/driver/node.go
+++ b/driver/node.go
@@ -40,57 +40,16 @@ func NewNodeService(
 }
 
 func (s *NodeService) NodeStageVolume(ctx context.Context, req *proto.NodeStageVolumeRequest) (*proto.NodeStageVolumeResponse, error) {
-	if req.VolumeId == "" {
-		return nil, status.Error(codes.InvalidArgument, "missing volume id")
-	}
-	if req.StagingTargetPath == "" {
-		return nil, status.Error(codes.InvalidArgument, "missing staging target path")
-	}
-	if req.VolumeCapability == nil {
-		return nil, status.Error(codes.InvalidArgument, "missing volume capability")
-	}
-
-	switch {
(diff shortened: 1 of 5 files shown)


batistein:
@samcday great work on all of these PR's! Is there a repo were I could already use it? And is there a helm chart anywhere already? P.S. Maybe this is something for your personal infrastructure project: https://github.com/syself/cluster-api-provider-hetzner

samcday:
Hi!

> Is there a repo were I could already use it?

At the moment, it's just a set of branches in this repo. The branch for this PR, `remove-node-staging`, includes the full series of commits from the earlier PRs.

> And is there a helm chart anywhere already?

Alas, no. With the changes made thus far and the plan to support more complicated deployments, a Helm chart would certainly be useful. Contributions welcome!

> https://github.com/syself/cluster-api-provider-hetzner

Thanks, for the link, I've been watching your excellent work on a Hetzner CAPI integration for a while (I [forked your work back in April last year!](https://github.com/samcday/cluster-api-provider-hcloud) ). For my own personal cluster I don't plan to complicate the bootstrapping beyond some Bash scripts, though ;)

PR #276, opened 2022-02-16
fix secret name and use current deployment version for example

alexanderkraemer:


LKaemmerling:
Hey @alexanderkraemer,

the referenced secret name is correct. If you use the master version (latest) the default secret is now `hcloud`. This is already done as preparation for the next major version, on which we currently work on.

PR #282, opened 2022-02-23
Separate server ID from location for controller

NotTheEvilOne:
This PR adds support to execute the CSI controller on a node not hosted on the Hetzner Cloud. Additionally it reduces required API calls in this case as the location is already known at execution.

It adds a new environment variable `HCLOUD_SERVER_OVERRIDE_LOCATION_NAME` used for the controller service to work properly even without a server ID.

This is useful in a Gardener deployment where the "csi-driver-controller" is not necessarily deployed on a node in the Hetzner Cloud offering but still manages volumes for nodes deployed on the cloud.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -30,14 +30,21 @@ func main() {
 	}
 
 	metadataClient := metadata.NewClient(metadata.WithInstrumentation(m.Registry()))
+	var location string
 
-	server, err := app.GetServer(logger, hcloudClient, metadataClient)
-	if err != nil {
-		level.Error(logger).Log(
-			"msg", "failed to fetch server",
-			"err", err,
-		)
-		os.Exit(1)
+	if s := os.Getenv("HCLOUD_SERVER_OVERRIDE_LOCATION_NAME"); s != "" {
+		location = s
+	} else {
+		server, err := app.GetServer(logger, hcloudClient, metadataClient)
+		if err != nil {
+			level.Error(logger).Log(
+				"msg", "failed to fetch server",
+				"err", err,
+			)
+			os.Exit(1)
+		}
+
+		location = server.Datacenter.Location.Name
 	}
 
 	volumeService := volumes.NewIdempotentService(
@@ -50,7 +57,7 @@ func main() {
 	controllerService := driver.NewControllerService(
 		log.With(logger, "component", "driver-controller-service"),
 		volumeService,
-		server.Datacenter.Location.Name,
+		location,
 	)
 	identityService := driver.NewIdentityService(
 		log.With(logger, "component", "driver-identity-service"),


NotTheEvilOne:
Is there any news regarding this repository @LKaemmerling or @fhofherr ?

apricote:
It looks like this location is only used once: https://github.com/hetznercloud/csi-driver/blob/c83b6cc77a7f890a40095cdfd762e001badd602c/driver/controller.go#L57-L63

Considering this, I wonder If we can just completely remove the code, as all volumes should have topology requirements attached by now (GA in Kubernetes since 1.17).

apricote:
> It looks like this location is only used once:
> 
> https://github.com/hetznercloud/csi-driver/blob/c83b6cc77a7f890a40095cdfd762e001badd602c/driver/controller.go#L57-L63
> 
> Considering this, I wonder If we can just completely remove the code, as all volumes should have topology requirements attached by now (GA in Kubernetes since 1.17).

While this is (IMO) technically allowed according to the CSI spec, the `kubernetes-csi/csi-test` sanity tests fail if we require the accessibilityRequirements to always be set.

I researched how other csi-drivers are doing this, and some also accept config options to override the default location (`digitalocean` and `gce-persistent-disk`).

cmd/controller/main.go
```diff
@@ -30,14 +30,21 @@ func main() {
...
-	if err != nil {
-		level.Error(logger).Log(
-			"msg", "failed to fetch server",
-			"err", err,
-		)
-		os.Exit(1)
+	if s := os.Getenv("HCLOUD_SERVER_OVERRIDE_LOCATION_NAME"); s != "" {
+		location = s
+	} else {
+		server, err := app.GetServer(logger, hcloudClient, metadataClient)
```
apricote, line 38: `metadataClient` is only used in this branch. Can you move the `metadataClient := metadata.NewClient()` line here?

cmd/controller/main.go
```diff
@@ -30,14 +30,21 @@ func main() {
...
+	var location string
 
-	server, err := app.GetServer(logger, hcloudClient, metadataClient)
-	if err != nil {
-		level.Error(logger).Log(
-			"msg", "failed to fetch server",
-			"err", err,
-		)
-		os.Exit(1)
+	if s := os.Getenv("HCLOUD_SERVER_OVERRIDE_LOCATION_NAME"); s != "" {
```
apricote, line 35: What do you think about calling this var `HCLOUD_VOLUME_DEFAULT_LOCATION` instead?

While this technically provides an override to the server location check, the returned value is used as the default location for volumes which have no location specified.

apricote (changes requested):

apricote:
Hey @NotTheEvilOne, 

I saw that you created 23technologies/hcloud-csi-driver#1 with the requested changes. Should we wait for that PR or just create a new one here with the contents?

NotTheEvilOne:
Good morning,

I guess we should just go ahead and take over the changes from my [user repository](https://github.com/hetznercloud/csi-driver/compare/main...NotTheEvilOne:hcloud-csi-driver:patch-1). Please let me know if you want me to create a new PR for it.

apricote:
> Good morning,
> 
> I guess we should just go ahead and take over the changes from my [user repository](https://github.com/hetznercloud/csi-driver/compare/main...NotTheEvilOne:hcloud-csi-driver:patch-1). Please let me know if you want me to create a new PR for it.

That would be great :)

NotTheEvilOne:
Superseded by #385.

apricote:
Replaced by #385.

PR #290, opened 2022-05-22
Fix secret name in README.md

SimonRosenau:
The csi-driver uses the secret `hcloud-csi` instead of `hcloud`

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ enabling you to use ReadWriteOnce Volumes within Kubernetes. Please note that th
    apiVersion: v1
    kind: Secret
    metadata:
-     name: hcloud
+     name: hcloud-csi
      namespace: kube-system
    stringData:
      token: YOURTOKEN


mvhirsch (approved):

LKaemmerling:
Hey @mvhirsch,

this is not completely correct. The master/main version already include changes for a v2.0 of the csi driver. With this major version increase, we also plan to unify the secret name with the cloud controller manager (so hcloud is correct).

mvhirsch:
> The master/main version [...]

@LKaemmerling you don't speek [about 1.6.0](https://github.com/hetznercloud/csi-driver/releases/tag/v1.6.0), do you? If so, the current `README.md` might be correct (according to `master`), but any user has troubles anyway (using a to-new documentation on old releases). Btw. the release version in `README.md` is still outdated (1.6.0, a.k.a. "latest" if I'm correct). I've just opened a PR for that: https://github.com/hetznercloud/csi-driver/pull/291

To avoid inconsistencies either _release a new version_ (which is in line with `README.md`) **or** _create a new default branch_ (say 1.6) to link the "old" `README.md`. New users will than see the correct secret-name according to 1.6.0. 

What do you think?

PR #293, opened 2022-05-28
Get the `serverLocation` by using the hcloud client instead of metadata client

omBratteng:
Because older VMs, or VMs created from a snapshot, does not have the `availability-zone` in the metadata service.
So this is a workaround by using the hcloud client to call the hcloud API, and mounts the HCLOUD_TOKEN to the hcloud-csi-node

Related issue https://github.com/hetznercloud/hcloud-go/issues/192?

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -1,10 +1,9 @@
 package main
 
 import (
-	"fmt"
+	"context"
 	"os"
 	"strconv"
-	"strings"
 
 	proto "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


omBratteng:
I ran the e2e tests in my fork, all green, https://github.com/omBratteng/hetznercloud-csi-driver/actions/runs/2399438268

hakman:
Is this a good idea from security point of view? Having `HCLOUD_TOKEN` on each node is not exactly a good practice.
Maybe use the hcloud API only when `HCLOUD_TOKEN` is available, otherwise continue using the metadata.

omBratteng:
In regards of rate limiting, not sure that would be reached. As it would only be ran at startup of each replica in the DaemonSet.
And in regards of security, I'm not sure. The secret is installed in the cluster and namespace already.

Ideally, the metadata would be available for all VMs, or, adding a new metadata key just saying which region it is in, as that's what we're really after. And I hope that wouldn't be something that would break backwards compatibility if it were added.

hakman:
If I recall correctly, a Secret is only sent to a node [if a pod on that node requires it](https://kubernetes.io/docs/concepts/configuration/secret/#information-security-for-secrets). Adding the token to `hcloud-csi-node` would expose the secret to all nodes.
You may also want to take a look at https://github.com/hetznercloud/csi-driver/pull/269, which you are mostly reverting.

omBratteng:
I don't disagree that this isn't the best solution. But the issue stems from not being able to access the `availability-zone` in the metadata endpoint if you have an older VM or create a VM from a snapshot.

I'll do a bit more deep dive into why we need the availabilty-zone, so far I haven't seen any actual use for it. Might be part of CSI spec

sphr2k:
> I'll do a bit more deep dive into why we need the availabilty-zone, so far I haven't seen any actual use for it. Might be part of CSI spec

It's required for [CSI Topology](https://kubernetes-csi.github.io/docs/topology.html) awareness and it's relevant for Hetzner Cloud, since volumes located in one location cannot be attached to a server in another location.

apricote:
Hello @omBratteng

the `availability-zone` metadata service field is set for all servers by now. If your server has been running for a long time, you need to restart it for the field to appear. The field is also set for all servers created from snapshot.

I am going to close this PR, as the functionality works as intended. If you still have issues, feel free to create a new issue to discuss your use case.

sphr2k:
@apricote Unfortunately, the `availability-zone` field is still not populated for at least three of our servers _after_ rebooting.

```
$ curl http://169.254.169.254/hetzner/v1/metadata/availability-zone
availability-zone not found
```

Should I create a new issue? Would it help if I sent you the server IDs?

apricote:
I am sorry for not updating this issue.

Between my message from last week and yesterday I found out that the `availability-zone` field is not actually set for all existing servers, and a simple reboot does not help. We are currently working out the details to fix this for existing servers. The release of v2.0.0 (that includes the change requiring this) is blocked until we find an acceptable solution that does not require users to create new servers.

apricote:
@sphr2k We updated the metadata for all existing servers. The `availablity-zone` field should now be available on all your servers without a reboot.

Issue #304, opened 2022-08-02
LUKS not functioning as intended


EternalDeiwos:
I've deployed the CSI driver from `master` to a Rancher (2.6.6) cluster (rke1 k8s 1.23) to try out LUKS (#279) and found that volumes are not being mapped correctly.

I set up the SC as follows:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: hcloud-volumes-encrypted
provisioner: csi.hetzner.cloud
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
allowVolumeExpansion: true
parameters:
  csi.storage.k8s.io/node-publish-secret-name: encryption-secret
  csi.storage.k8s.io/node-publish-secret-namespace: kube-system
```

Logs on the CSI provisioner and controller show normal creation, attaching, detaching, and deletion of the volume &mdash; nothing mentioning LUKS or encryption. The volume is accessible and can be used as normal within the cluster via PVC, or manually mounted on another node.

EternalDeiwos:
It seems that the latest changes to master have not been published into a Docker image yet.

EternalDeiwos:
With the updated Docker image this works as intended

PR #305, opened 2022-08-09
ci: publish unstable docker image from main

EternalDeiwos:
This should fix publishing unstable docker images and allow us to experiment with new features (e.g. #304)


LKaemmerling:
Hey @EternalDeiwos,

thank you! Good catch this was overseen.

PR #309, opened 2022-08-23
Build the Docker image on any platform

hakman:
The resulting Docker image should have the same platform, no matter where it's built.
This change will make sure both the binaries and the base image are always `linux/amd64`.

I also updated the Alpline Linux image to 3.16, as 3.13 is approaching end of support.


hakman:
@apricote Was wondering if this approach if of any use in general.
My use case was to be able to be build on a M1 Mac.

apricote:
> @apricote Was wondering if this approach if of any use in general. My use case was to be able to be build on a M1 Mac.

Thanks for the contribution! It certainly is in my opinion, we should make sure that anyone can contribute to the project, whether they use ARM or x86_64.

PR #311, opened 2022-08-30
master --> main

guettli:
Fix typo, where central branch was still called "master".

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,11 +132,11 @@ related only to an unsupported version.
 
 | Kubernetes |    CSI Driver |                                                                                   Deployment File |
 |------------|--------------:|--------------------------------------------------------------------------------------------------:|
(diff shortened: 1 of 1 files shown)


jlgeering:
while we are at it ... should we also fix the download url: from `https://raw.githubusercontent.com/hetznercloud/csi-driver/v1.6.0/deploy/kubernetes/hcloud-csi.yml` to `https://raw.githubusercontent.com/hetznercloud/csi-driver/main/deploy/kubernetes/hcloud-csi.yml`

@LKaemmerling what do you think?

LKaemmerling:
@guettli good catch thank you!

@jlgeering yes and no. We don't recommend using the main/development branch. It currently contains a lot of breaking changes that we want to continue before releasing it to the wild.

jlgeering:
ok, sure, but ... it's just a bit confusing now, as the table in the readme says we should be using main for kubernetes 1.23 and 1.24 but then links to the "old" version

PR #317, opened 2022-09-22
StorageClass has cluster scope

jlgeering:


apricote (approved):

apricote (changes requested):
Thanks for your contribution @jlgeering!

Can you also remove the namespace attribute from the StorageClass in `deploy/kubernetes/core/storageclass.yaml`? This is used to generate the file `deploy/kubernetes/hcloud-csi.yml`

apricote (approved):
Looking great! Thanks again for the contribution.

PR #322, opened 2022-11-11
test: fix integration tests relying on specific byte amounts

apricote:
The integration tests for the volume resize feature rely on specific byte amounts that a ext4 filesystem on a fake block device has. The specific amounts recently changed, probably through a new version of mkfs.ext4.

This amount equals the values from our GitHub Actions pipeline, as well as my local machine.

The amounts might need to be changed again in the future if the tests start failing again, we might consider removing tests for the specific amounts, and just check that the resized number is larger than the initial if this happens often.

### DIFF
--- a/integrationtests/volumes_test.go
+++ b/integrationtests/volumes_test.go
@@ -149,8 +149,8 @@ func TestVolumeResize(t *testing.T) {
 		initialSize int
 		finalSize   int
 	}{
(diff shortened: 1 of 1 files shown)


choffmeister:
@apricote Interesting. Didn't think of that back then when I wrote that code. Another idea to relax the test is to just look at the ratio between old and new size, round it and expect it to equal 2.

PR #342, opened 2022-12-08
encrypted storageclass had the wrong names. reference to cryptsetup

JustinGuese:
the referenced secret name and namespace was wrong.
Also I added a remark for the nodes to have the cryptsetup package installed, as not all nodes have it by default.
I have tested the name change of the secret in my cluster and can confirm that encryption works


apricote (approved):
Thanks a lot for the contribution!

Issue #278, opened 2022-02-18
CSI Mount Error on Hetzner


martinbe1io:
Hi all, 

hope somebody has a hint for me as I already invested hours without a solution, in the beginning everything with my kuberone cluster on Hetzner worked as expected. auto provisioning of block storage over the csi interface is exactly as it should. Since days across the different namespaces errors occur at initial pod starting after provisioning via helm

**the error message at pod startup**
```
MountVolume.MountDevice failed for volume "pvc-a0a20263-9462-4c3e-9ebb-be45b92da7f4" : rpc error: code = Internal desc = failed to stage volume: format of disk "/dev/disk/by-id/scsi-0HC_Volume_16943979" failed: type:("ext4") target:("/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-a0a20263-9462-4c3e-9ebb-be45b92da7f4/globalmount") options:("defaults") errcode:(exit status 1) output:(mke2fs 1.45.7 (28-Jan-2021) The file /dev/disk/by-id/scsi-0HC_Volume_16943979 does not exist and no size was specified. )
```

```
Unable to attach or mount volumes: unmounted volumes=[redis-pvc], unattached volumes=[redis-pvc default-token-g7mjq]: timed out waiting for the condition
```

**parts of the deployment file**

```
spec:
  volumes:
    - name: redis-pvc
      persistentVolumeClaim:
        claimName: redis-pvc-chirpstack-redis-0
    - name: default-token-g7mjq
      secret:
        secretName: default-token-g7mjq
        defaultMode: 420


 volumeMounts:
        - name: redis-pvc
          mountPath: /data
        - name: default-token-g7mjq
          readOnly: true
          mountPath: /var/run/secrets/kubernetes.io/serviceaccount
```

pvc is like this

```

apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: redis-pvc-chirpstack-redis-0
  namespace: chirpstack
  selfLink: >-
    /api/v1/namespaces/chirpstack/persistentvolumeclaims/redis-pvc-chirpstack-redis-0
  uid: 70520ae2-b99f-4d7f-a625-5e97d1748dd9
  resourceVersion: '2367972'
  creationTimestamp: '2022-02-15T17:44:04Z'
  labels:
    app: redis
    release: chirpstack
  annotations:
    pv.kubernetes.io/bind-completed: 'yes'
    pv.kubernetes.io/bound-by-controller: 'yes'
[...]

PR #343, opened 2022-12-09
Add extra checks for volume attachment

alam0rt:
Volume publish can succeed even if the volume is not yet attached or mounted to the underlying Linux server.

https://github.com/hetznercloud/csi-driver/issues/278

The `VolumeAttachement` is meant to contain the mount path:
```
  status:
    attached: true
    attachmentMetadata:
      devicePath: /dev/disk/by-id/scsi-0HC_Volume_20867585
```

but sometimes it does not, meaning that the `ControllerPublishVolumeResponse` did not contain it either.

```
  status:
    attached: true
```

It seems that since this request is only issued once, if the mount path isn't set, it won't be set until the `VolumeAttachment` is deleted and `ControllerPublishVolume` is called again.

According to the [API](https://docs.hetzner.cloud/#volumes-get-a-volume), the `server` field of the  `GET /volume/{id}` is nullable, meaning it won't be set if the volume isn't yet attached.

This PR adds a check to see if the volume is actually attached or not.

## Investigation

```
hcloud-csi-driver level=error ts=2022-12-09T06:24:54.932343015Z component=grpc-server msg="handler failed" err="rpc error: code = Internal desc = failed to publish volume: exit status 1\nmke2fs 1.46.4 (18-Aug-2021)\nThe file  does not exist and no size was specified.\n"
```

Looking at the error, we can see there is an extra space after file in `file  does not...`. Additionally, the log line `component=linux-mount-service msg="formatting disk" disk= fstype=ext4` reveals the `disk` variable is empty. It seems the CSI isn't getting the device path.

I had a quick look at the CSI spec, it seems the [`NodePublishVolume()` request](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume) is the thing which is failing. Specifically it looks like the path isn't in the map here https://github.com/hetznercloud/csi-driver/blob/main/driver/node.go#L60.

That `map[string]string` is apparently set by `ControllerPublishVolume()`:

```
message NodePublishVolumeRequest {
  // The ID of the volume to publish. This field is REQUIRED.
  string volume_id = 1;

  // The CO SHALL set this field to the value returned by
  // `ControllerPublishVolume` if the corresponding Controller Plugin
  // has `PUBLISH_UNPUBLISH_VOLUME` controller capability, and SHALL be
  // left unset if the corresponding Controller Plugin does not have
  // this capability. This is an OPTIONAL field.
  map<string, string> publish_context = 2;
```

This leads to https://github.com/hetznercloud/csi-driver/blob/main/driver/controller.go#L176-L190 which in turn calls https://github.com/hetznercloud/csi-driver/blob/6beac22cec3523a338241ead8a9f6f394c15c666/api/volume.go#L83-L101.

### DIFF
--- a/csi/volume.go
+++ b/csi/volume.go
@@ -13,3 +13,15 @@ type Volume struct {
 func (v Volume) SizeBytes() int64 {
 	return int64(v.Size) * 1024 * 1024 * 1024
 }
+
+// IsAttached returns true if the volume is not attached to
+// a server.
+func (v Volume) IsAttached() bool {
+	return v.Server != nil
+}
+
+// IsMounted returns true if the volume has been mounted to a
+// Linux server and has a mount path.
+func (v Volume) IsMounted() bool {
+	return v.LinuxDevice != ""
+}
--- a/driver/controller.go
+++ b/driver/controller.go
@@ -182,6 +182,9 @@ func (s *ControllerService) ControllerPublishVolume(ctx context.Context, req *pr
 			return nil, status.Error(codes.Internal, fmt.Sprintf("failed to get volume: %s", err))
 		}
 	}
+	if !volume.IsMounted() || !volume.IsAttached() {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("volume not yet attached: mounted=%t, attached=%t", volume.IsMounted(), volume.IsAttached()))
+	}
 
 	resp := &proto.ControllerPublishVolumeResponse{
 		PublishContext: map[string]string{


apricote:
The unit tests seem to encounter the bug, I am currently investigating internally whether this is expected and we need to implement a workaround.

alam0rt:
> The unit tests seem to encounter the bug, I am currently investigating internally whether this is expected and we need to implement a workaround.

perhaps the "fix" i've implemented should be in the hcloud-go library? Not sure

Unsure where-else the `.status.attachementMetadata` would be getting reset from... /shrug

apricote:
> > The unit tests seem to encounter the bug, I am currently investigating internally whether this is expected and we need to implement a workaround.
> 
> perhaps the "fix" i've implemented should be in the hcloud-go library? Not sure
> 
> Unsure where-else the `.status.attachementMetadata` would be getting reset from... /shrug

I think I found the scenario where the empty attachmentMetadata is coming from: VolumeAttachments created with old csi-driver. See my comment: https://github.com/hetznercloud/csi-driver/issues/278#issuecomment-1351652816

Based on this, I do not think that the checks from this PR are required. Instead we should verify that `devicePath` is set before trying to mount the volume, and otherwise tell users that they need to re-attach the volume. I created PR #344 for this new check.

---

The unit tests actually failed because we mock the API during the tests and this mock did not set the volume, causing the new checks to fail.

apricote:
Closing this MR as the checks are not necessary: https://github.com/hetznercloud/csi-driver/pull/343#issuecomment-1351711045

Issue #349, opened 2022-12-22
Volume Formatting breaks Block mode in v2


m4eba:
Every volume gets formatted even if in block mode.

timbrd:
This might be the cause why I am not able to provision a rook ceph cluster anymore.
When initializing the osds, rook does not find any valid block devices:

````
provision 2022-12-23 17:34:24.432561 I cephosd: skipping device "/mnt/set1-data-1tl48r" because it contains a filesystem "ext4"
provision 2022-12-23 17:34:24.461665 I cephosd: configuring osd devices: {"Entries":{}}                                                                                        
provision 2022-12-23 17:34:24.461685 I cephosd: no new devices to configure. returning devices already configured with ceph-volume.
````

The pvc configured is in block mode:

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  [...]
  name: set1-data-1tl48r
  namespace: rook-ceph
  [...]
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 90Gi
  storageClassName: hcloud-volumes
  volumeMode: Block
  volumeName: pvc-00bbd41c-548a-4d6d-a33d-05e0beba2584
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 90Gi
  phase: Bound
```

timbrd:
When I switch back to csi driver 1.6.0, I was able to provision a ceph cluster.

m4eba:
I also run into this playing around with rook. What works for me is a simple if statement for the code that formats the volume: https://github.com/m4eba/csi-driver/commit/c3c8173013f306a8765c56fae5a9638c0b2dd17b

Privatecoder:
this also breaks re-attaching LUKS encrypted volumes

apricote:
Hey all, thanks for the report. I am currently working on this, the fix is ready but I want some proper tests around this to make sure that it will not break again in the future.

apricote:
Fixed by #353, will be released in the next patch release.

Privatecoder:
@apricote v2 (any version) does throw the error. Downgrading to 1.6.0 fixes it, however I‘d like to use a more recent version and not lose the pvcs.

apricote:
It sounds like you setup the volumes with v1.6, but that version does not include support for encrypted volumes. Encrypted volumes were first released in v2.0.0. If your volumes are actually LUKS encrypted, attaching them with v1.6 should fail.

[...]

PR #279, opened 2022-02-18
Add support for volume encryption with cryptsetup and LUKS

choffmeister:
This PR adds full volume encryption. In short, one just has to add a new storage class referencing an encryption passphrase:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: encryption-secret
  namespace: kube-system
stringData:
  encryption-passphrase: foobar

--- 

apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: hcloud-volumes-encrypted
provisioner: csi.hetzner.cloud
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
allowVolumeExpansion: true
parameters:
  csi.storage.k8s.io/node-publish-secret-name: encryption-secret
  csi.storage.k8s.io/node-publish-secret-namespace: kube-system
```

~~What is still missing is adjusting the e2e tests to also cover this (or do they already cover that case)?~~

Solves #40 and #163.

Note: ~~I did not make the PR against the latest `master`, as I had issues with it (the availability zone request [here](https://github.com/hetznercloud/csi-driver/blob/master/cmd/node/main.go#L33) returns HTTP 404 on my Hetzner test K8s cluster and hence the csi-node wouldn't start up). I also created a PR for that issue: https://github.com/hetznercloud/hcloud-go/issues/192.
If this feature would be accepted in general, I am happy to update the PR to fit into the master.~~

PR #353, opened 2023-01-12
fix: Block volumes are formatted #349

apricote:
A recently introduced change (#279) caused block volumes to be formatted with ext4. This happened because setting the default for the FS type was moved from the non-Block Volume branch into the general code.

This PR reverts the move and adds a test to make sure that block storage is not accidentally formatted again in the future.

### DIFF
--- a/integrationtests/docker.go
+++ b/integrationtests/docker.go
@@ -48,7 +48,7 @@ func runCmdWithStdin(stdin string, name string, args ...string) (string, error) 
 	outputBytes, err := cmd.CombinedOutput()
 	output := string(outputBytes)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


LKaemmerling (approved):

toabi:
@apricote is there a date planned for releasing this fix?

apricote:
> @apricote is there a date planned for releasing this fix?

I plan on releasing a new version today.

toabi:
@apricote tried it and it is indeed fixed, just manual deletion of volumeattachments is sometimes necessary.

By the way, I noticed the README was not updated with links to the latest version.

PR #358, opened 2023-01-13
Remove explicit image pull policies

choffmeister:
If image pull policy is not specified, then latest images implicitly have the pull policy Always. All other sig-storage images are versioned and hence have IfNotPresent by default.

### DIFF
--- a/deploy/kubernetes/controller/deployment.yaml
+++ b/deploy/kubernetes/controller/deployment.yaml
@@ -37,7 +37,6 @@ spec:
           mountPath: /run/csi
       - name: hcloud-csi-driver
         image: hetznercloud/hcloud-csi-driver:latest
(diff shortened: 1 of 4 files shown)


apricote:
Is there a reason why you prefer `IfNotPresent` over `Always`?

choffmeister:
@apricote When using fixed tags `IfNotPresent` is better. When using specific tag versions a once uploaded image should never be updated in place and hence there is no need to refetch it.

Especially for infrastructure critical workloads like a CSI. Imaging that k8s.gcr.io is not available and one of the CSI pods crashes, then they cannot be restarted until the registry is back up which can lead to downtime of workloads.

choffmeister:
@apricote Bump

Issue #361, opened 2023-01-13
ci: E2E tests fail because gcc is not found


apricote:
Currently all e2e test pipelines fail with

```
go: downloading github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578
# runtime/cgo
cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
FAIL	github.com/hetznercloud/csi-driver/e2etests [build failed]
FAIL
```

See for example:
- https://github.com/hetznercloud/csi-driver/actions/runs/3910151870/jobs/6682010887#step:7:68
- https://github.com/hetznercloud/csi-driver/actions/runs/3910120760/jobs/6681944181#step:7:68

apricote:
We self host the github runners, and use [github.com/actions/actions-runner-controller](https://github.com/actions/actions-runner-controller/) for this. By default, this starts runners with the image `summerwind/actions-runner:latest`, but the `latest` tag is deprecated and might contain ubuntu-20.04 or ubuntu-22.04, I suspect this might be the cause of the sudden change, as the `latest` tag was changed recently and points to 22.04 atm.

apricote:
Using `ubuntu-20.04` for the ci image fixed the problem.

PR #362, opened 2023-01-13
ci: disable CGO in e2e pipeline

apricote:
We do not require CGO for our e2e pipeline, but somehow it started being used on 2023-01-13.


apricote:
Not required for now, we instead reverted back to using ubuntu-20.04 for the GHA image. See #361

PR #363, opened 2023-01-18
 ci: setup release-please to automate new releases

apricote:
See https://github.com/hetznercloud/hcloud-go/pull/223 for detailed explanation


apricote:
This effort is paused until we move the `hcloud-csi.yml` with defined tags to the release assets.

Issue #346, opened 2022-12-15
Rate limit exceeded creating 6 or more volumes


diegoparrilla:
Hi all,

I have found this error twice, when creating a Kafka and a Redis cluster with three pods and  two volumes each. This error makes the deployment to fail after several minutes:

Describing the failing pod (all other 5 pods worked):
```
kubectl describe pod redis-cluster-follower-1 --namespace redis-operator
```

returns:
```
Events:
  Type     Reason              Age                   From                     Message
  ----     ------              ----                  ----                     -------
  Normal   Scheduled           9m28s                 default-scheduler        Successfully assigned redis-operator/redis-cluster-follower-1 to node-2-default
  Warning  FailedAttachVolume  67s (x12 over 9m28s)  attachdetach-controller  AttachVolume.Attach failed for volume "pvc-f6490646-0aee-413d-af34-fc52137270f1" : rpc error: code = Internal desc = failed to publish volume: limit of 3600 requests per hour reached (rate_limit_exceeded)
  Warning  FailedMount         38s (x4 over 7m26s)   kubelet                  Unable to attach or mount volumes: unmounted volumes=[redis-cluster-follower], unattached volumes=[redis-cluster-follower kube-api-access-4tfh2]: timed out waiting for the condition
```

As I said I found the same problem creating a Kafka cluster with six pods and six volumes. It seems that the rate limit of the Hetzner API is too low for the number of requests performed by the `csi-driver`. So my questions are:

1. Is there a way to reduce the number of requests?
2. Any way to increase the rate limit of the API?
3. Any other suggested workaround will be welcomed!

Thanks!

apricote:
Hey @diegoparrilla,

the API rate limit is calculated on a per-project basis. In general I do not expect the `csi-driver` to reach the rate limit. Did you make a lot of other requests in the same project?

To monitor the number of API requests made by the `csi-driver`, you can configure Prometheus (or other metrics tools) to scrape the metrics from the endpoint `:9189` on all csi-driver pods. This will include metrics with the prefix `hcloud_api_` which provide information about all requests made to the API. Using these metrics, you can estimate how many requests are made per time frame, and if this is too many.

To answer your questions:

> 1. Is there a way to reduce the number of requests?

No, but the scenario you described should not exceed the rate limit.

> 2. Any way to increase the rate limit of the API?

You can message the [support](https://console.hetzner.cloud/support) for this.

> 3. Any other suggested workaround will be welcomed!

[...]

Issue #369, opened 2023-01-24
feat: Publish Helm Chart


apricote:
Some use-cases require deeper customization of our deployment manifests. We should publish and provide a Helm chart for these users. The "simple" deployment manifests can then be generated from the Helm chart, by building it with the default values.

Example usecases that require deep customization:
- [x] non-standard kubelet config path, e.g. in k0s: #357 (in #516)
- [x] (non-)Default StorageClass https://github.com/hetznercloud/csi-driver/issues/369#issuecomment-1410005390 (in #500)
- [x] Image Pull Policy #358 (in #500)
- [x] Deploy `ServiceMonitor` and Grafana dashboard https://github.com/hetznercloud/csi-driver/pull/378#issuecomment-1440418738 (ServiceMonitor in #500, Grafana Dashboard moved to https://github.com/hetznercloud/csi-driver/issues/557)

toabi:
That's a good plan. Another useful option for this would be to install it as non-default storage class.

batistein:
@apricote you could use our helm chart as a starting point if you want. See: https://github.com/syself/charts/tree/main/charts/csi-hcloud

If I have some time in the next couple of weeks I will update it to the v2 version of the csi-driver.

apricote:
The base Helm Chart is now in the repo and the remaining features from the list are available for pickup if anyone is interested :)

apricote:
I opened https://github.com/hetznercloud/csi-driver/issues/557 for deploying the grafana dashboard in the chart. Closing this issue.

PR #378, opened 2023-02-14
feat: add Grafana dashboard and ServiceMonitors

apricote:
Add a default dashboard to visualize the exposed metrics as well as `ServiceMonitors` and a guide on how to set everything up.

Related to #346

### DIFF
--- a/deploy/kubernetes/controller/service.yaml
+++ b/deploy/kubernetes/controller/service.yaml
@@ -4,7 +4,7 @@ metadata:
   name: hcloud-csi-controller-metrics
   namespace: kube-system
   labels:
-    app: hcloud-csi
+    app: hcloud-csi-controller
 spec:
   selector:
     app: hcloud-csi-controller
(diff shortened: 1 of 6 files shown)


samcday (approved):

splattner:
this should also be added to the kustomization.yaml

apricote:
Hey @splattner, thanks for the suggestion :) I was thinking about that, but wanted to keep the existing `kustomization.yaml` backwards compatible. The problems with that approach are:

- If we add the `ServiceMonitors` to the main `kustomization.yaml`, it will require `prometheus-operator` to be installed beforehand (or at least its CRDs) and fail for clusters that dont have the operator
- There is no standard way to deploy Grafana dashboards, we could make a ConfigMap out of it though

We did add a new `kustomization.yaml` that you can use to install the ServiceMonitors: `deploy/kubernetes/service-monitor/kustomization.yaml`

I also plan on adding this to an upcoming Helm Chart, see #369

PR #396, opened 2023-03-24
Add volumeattachment script macos compatibility

laurigates:
The macos commands only accept the short form of the options.


apricote (approved):

apricote:
Thanks for the contribution @laurigates!

PR #394, opened 2023-03-21
Fix typo in Script: Remove quotes.

guettli:
With quotes, all IDs are in `$VOLUME_ATTACHMENT`.

PR #397, opened 2023-03-24
Fix volumeattachment array

laurigates:
This change was required to have the array handled correctly at least on the bash version included in macos.


apricote:
This at least kind of reverts #394, and I am pretty sure I had this exact code in the script at some point.

I dont like coding with bash... Going to take a detailed look at this again next week.

apricote (approved):

apricote:
Thanks for the fix @laurigates :)

PR #429, opened 2023-07-05
ci: replace deprecated ::set-output call

jooola:
Replace the deprecated `::set-output .*::` call using environment variables.


apricote (approved):
CI fails for unrelated reasons. We should remember to publish a release candidate for the next release to validate these changes.

Issue #331, opened 2022-11-21
feat: Support for Nomad


apricote:
Nomad supports the CSI specification, and our provider works with it. At the moment we only test against and provide documentation for Kubernetes. Nomad users have to figure out how to deploy the driver themselves (which is arguably way easier than the kubernetes deployment).

This has caused issues, where changes we made to our driver and the Kubernetes deployment were not communicated to Nomad users, whose deployment broke as a result of this. See #330.

To support Nomad we would need to implement the following tasks:

- [ ] Provide Nomad Job definitions
- [ ] Document how to install the driver in Nomad
- [ ] Test Nomad support based on the Job definitions to make sure that we do not break it

apricote:
> I would like to contribute the docs and jobs but I am missing a doc structure entrypoint.

Very nice :) I have started to refactor the existing docs to differentiate what is for Kubernetes vs general stuff.

apricote:
@resmo I moved the existing docs & e2e tests into sub dirs to make room for Nomand stuff :) You can check out the changes in #410.

resmo:
PR #467

apricote:
Once we have some automated tests we can remove the "Not Officially Supported" warning from the README.

resmo:
closed by #467

apricote:
For now, I am really happy that we test the basic functionality. Happy to accept further test cases from the community, similar to the kubernetes CSI test suite.

PR #467, opened 2023-07-26
docs: add nomad docs

resmo:
See #331

### DIFF
--- a/docs/nomad/README.md
+++ b/docs/nomad/README.md
@@ -70,7 +70,7 @@ job "hcloud-csi-controller" {
         data        = <<EOH
 HCLOUD_TOKEN="{{ with nomadVar "secrets/hcloud" }}{{ .hcloud_token }}{{ end }}"
 EOH
-        destination = "secrets/hcloud-token.env"
+        destination = "${NOMAD_SECRETS_DIR}/hcloud-token.env"
         env         = true
       }
 
@@ -118,7 +118,7 @@ job "hcloud-csi-node" {
         data        = <<EOH
 HCLOUD_TOKEN="{{ with nomadVar "secrets/hcloud" }}{{ .hcloud_token }}{{ end }}"
 EOH
-        destination = "secrets/hcloud-token.env"
+        destination = "${NOMAD_SECRETS_DIR}/hcloud-token.env"
         env         = true
       }
 
--- a/docs/nomad/README.md
+++ b/docs/nomad/README.md
@@ -3,6 +3,7 @@
 ## Preconditions
 
 - Nomad >= 1.4.x cluster installed following the [Nomad Reference Architecture for production deployments](https://developer.hashicorp.com/nomad/tutorials/enterprise/production-reference-architecture-vm-with-consul). The setup was tested on Nomad Community, version 1.5.x.
+- The cluster nodes need to have the `docker` driver installed & configured with [`allow_privileged = true`](https://developer.hashicorp.com/nomad/docs/drivers/docker#allow_privileged).
 - The HCL resources are meant to be executed on a machine having nomad installed (with access to the Nomad API).
 
 ## Getting Started
--- a/docs/nomad/README.md
+++ b/docs/nomad/README.md
@@ -181,7 +181,7 @@ nomad volume create db-vol.hcl
 
 1. Create a Job definition
 
-In the following exmaple is shown, how we can mount the volume in a Docker Nomad job definition:
+The following example describes how to mount the volume in a Docker Nomad job definition:
 
 ```hcl
 # file: mariadb.nomad
--- a/docs/nomad/README.md
+++ b/docs/nomad/README.md
@@ -21,7 +21,7 @@ export HCLOUD_TOKEN="..."
 nomad var put secrets/hcloud hcloud_token=$HCLOUD_TOKEN
 ```
 
-3. Create a CSI Controller Job
+3. Create a CSI Controller Job:
 
 ```hcl
 # file: hcloud-csi-controller.hcl
--- a/docs/nomad/README.md
+++ b/docs/nomad/README.md
@@ -160,6 +160,9 @@ nomad plugin status
 
 Create a file `db-vol.hcl` for the volume resource:
 
+> [!NOTE]
+> See [Nomad Volume Specification](https://developer.hashicorp.com/nomad/docs/other-specifications/volume) for more information.
+
 ```
 # file: db-vol.hcl
 
@@ -169,6 +172,9 @@ name      = "db-vol"
 namespace = "default"
 plugin_id = "csi.hetzner.cloud"
 
+# Default minimum capacity for Hetzner Cloud is 10G
+capacity_min = "10G"
+
 capability {
   access_mode     = "single-node-writer"
   attachment_mode = "file-system"
@@ -177,6 +183,19 @@ capability {
 mount_options {
   fs_type     = "ext4"
   mount_flags = ["discard", "defaults"]
+}
+```
+> [!IMPORTANT]
+> The volume will be created in the same Hetzner Cloud Location as the controller is deployed into.
+
+To define the Hetzner Cloud Location (CLI: `hcloud location list`) you would like to create the volume into, append the following snippet into the volume resource definition:
+
+```
+topology_request {
+  required {
+   # Use your desired location name here
+    topology { segments { "csi.hetzner.cloud/location" = "fsn1" } }
+  }
 }
 ```
 
--- a/docs/nomad/README.md
+++ b/docs/nomad/README.md
@@ -213,7 +213,7 @@ nomad volume create db-vol.hcl
 
 1. Create a Job definition:
 
-The following example describes how to mount the volume in a Docker Nomad job definition:
+The following example describes how to mount the volume in a Docker Nomad job definition (especially see the parts commented with `### THIS!`):
 
 ```hcl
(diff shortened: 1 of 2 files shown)


docs/nomad/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+2. Deploy CSI Controller
+
+```hcl
+# hcloud-csi-controller.hcl
+
+job "hcloud-csi-controller" {
+  datacenters = ["dc1"]
+  namespace   = "default"
+  type        = "system"
```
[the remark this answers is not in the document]
resmo: no, it shouldn't be a system job,this is a mistake

resmo:
Ready for review

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+> [!NOTE]
+> Consider using HashiCorp Vault for secrets management, see https://developer.hashicorp.com/nomad/docs/job-specification/template#vault-kv-api-v2
+
+```sh
+export HCLOUD_TOKEN="..."
+nomad var put secrets/hcloud hcloud_token=$HCLOUD_TOKEN
+```
+
+3. Create a CSI Controller Job
```
jooola, line 24:
```suggestion
3. Create a CSI Controller Job:
```

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+      resources {
+        cpu    = 100
+        memory = 64
+      }
+    }
+  }
+}
+```
+
+4. Create a CSI Node Job
```
jooola, line 87:
```suggestion
4. Create a CSI Node Job:
```

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+      resources {
+        cpu    = 100
+        memory = 64
+      }
+    }
+  }
+}
+```
+
+5. Deploy Jobs
```
jooola, line 135:
```suggestion
5. Deploy the Jobs:
```

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+```sh
+nomad job run hcloud-csi-controller.hcl
+nomad job run hcloud-csi-node.hcl
+
+# Check status
```
jooola, line 141: Could you extract this in an extra step, please? We want to make sure folks actually do the check, so having a dedicated step makes it more explicit.

e.g. `6. Check the plugin status:`

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+```hcl
+# file: hcloud-csi-controller.hcl
+
+job "hcloud-csi-controller" {
+  datacenters = ["dc1"]
+  namespace   = "default"
+  type        = "service"
+
+  group "controller" {
+    count = 2
```
jamil7, line 35: Why are 2 controllers needed? 🤔

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+2. Create a Volume
+
+```sh
+nomad volume create db-vol.hcl
+```
+
+> [!HINT]
+>  The hcloud cli is a convient way to verify if the volume was created: `hcloud volume list`.
```
jooola, line 178:
```suggestion
>  The hcloud cli provides a convient way to verify if the volume was created: `hcloud volume list`.
```

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+  attachment_mode = "file-system"
+}
+
+mount_options {
+  fs_type     = "ext4"
+  mount_flags = ["discard", "defaults"]
+}
+```
+
+2. Create a Volume
```
jooola, line 171:
```suggestion
2. Create a Volume:
```

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+> [!HINT]
+>  The hcloud cli is a convient way to verify if the volume was created: `hcloud volume list`.
+
+### Make use of the Volume
+
+1. Create a Job definition
+
+In the following exmaple is shown, how we can mount the volume in a Docker Nomad job definition:
```
jooola, line 184:
```suggestion
The following example describes how to mount the volume in a Docker Nomad job definition:
```

docs/nomad/README.md
```diff
@@ -0,0 +1,250 @@
...
+        cpu    = 300
+        memory = 256
+      }
+
+    }
+  }
+}
+```
+
+2. Create the Job
```
jooola, line 246:
```suggestion
2. Create the Job:
```

jooola (changes requested):

docs/nomad/README.md
resmo: The thinking behind this is, one thing we want to make sure that controller is always available. If the node goes down the controller is running on, a volume won't be able to be migrated, right?

That is why we defined 2 of the controllers to run and also defined:

```
    constraint {
      distinct_hosts = true
    }
```

To ensure, controller instances run on different nodes.

jooola: If having 2 controllers is a good idea (I'll let others judge if this is valid) in this documentation, maybe you should add a comment above the field with your explanation ? If one person already raised a question about this, maybe others will.

apricote: The csi-controller is a stateless gRPC service. There should be no issue running two of them, as long as the caller makes sure to not send duplicate requests (and even those should be handled properly). I do not know if Nomad utilizes multiple controllers.


---

If one deploys this to a single node Nomad cluster, will the Job show an error because only a single task could be created?

resmo: > If having 2 controllers is a good idea (I'll let others judge if this is valid) in this documentation, maybe you should add a comment above the field with your explanation ? If one person already raised a question about this, maybe others will.

makes sense. 

We run this in prod and works as expected. I remember a "pressure" on our cluster when running only one instance of controller in "node down running csi controller" situation. 

> If one deploys this to a single node Nomad cluster, will the Job show an error because only a single task could be created?

I am not sure about this, I would say that it would work with a warning. However, single node clusters are considered experimental for testing. We refer to HashiCorp's "Best Practice for Production" with 3 master nodes. We should point that out in the doc.

docs/nomad/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+capability {
+  access_mode     = "single-node-writer"
+  attachment_mode = "file-system"
+}
+
+mount_options {
+  fs_type     = "ext4"
+  mount_flags = ["discard", "defaults"]
+}
```
apricote, line 180: Can we add a note here that this will create the volume in the Hetzner Cloud Location where the server with the controller is running?

If another location is required, users will need to specify [topology_request](https://developer.hashicorp.com/nomad/docs/other-specifications/volume/topology_request) (not tested):

```hcl
topology_request {
  required {
   # Use your desired location name here
    topology { segments { "csi.hetzner.cloud/location" = "fsn1" } } 
  }
}
```

docs/nomad/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+    }
+  }
+}
+```
+
+2. Create the Job:
+
+```sh
+nomad job run mardiadb.nomad
```
apricote, line 261:
```suggestion
nomad job run mariadb.nomad
```

Small typo, you also named all other files `.hcl`, can we make this consistent?

apricote (commented):
Nice and exhaustive guide for using the CSI driver from Nomad :) Going to also request a review from a coworker using Nomad & the csi-driver privately.

docs/nomad/README.md
resmo: Sure, let's add a note, I am also okay to add a topology request example (never had that use-case though), we should problaby also add a link to https://developer.hashicorp.com/nomad/docs/other-specifications/volume so users find specific docs to volume handling in nomad quicker.

apricote (commented):
Some further comments from my coworker

docs/nomad/README.md
```diff
@@ -0,0 +1,262 @@
...
+        MARIADB_ROOT_PASSWORD = "<...>"
+        MARIADB_DATABASE      = "<...>"
+        MARIADB_USER          = "<...>"
+        MARIADB_PASSWORD      = "<...>"
+      }
+
+      service {
+        name = "db"
+        port = "mariadb"
+      }
```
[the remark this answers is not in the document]
resmo: good point, since Nomad has a "lightweight" native service discovery (https://www.hashicorp.com/blog/nomad-service-discovery), let's use this instead and add a note about consul.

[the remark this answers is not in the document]
resmo: see 48a51c65aac71c5f231fd27c23873f50f1543f20

apricote (approved):
All looking good from my POV :) Going to leave this open for another day in case one of the other two reviewers still have open comments.

Thank you very much for writing and testing this @resmo!

apricote:
> common problem on GH, my PR doesn't have access to your secrets, tests may fail. ;)

Yea, but I am confident that a markdown document won't break our e2e tests ;)

apricote:
One thing I forgot, but I can also add it right before merging, it would be nice to link to the new docs from our readme: https://github.com/hetznercloud/csi-driver#getting-started

apricote (approved):

kaspergrubbe:
Thank you so much for your work on this <3!

PR #478, opened 2023-08-24
ci: setup release-please

apricote:
The current release process for csi-driver creates and pushes a new commit and then moves the tag to that commit. This introduces the standard release-please workflow we have in most of our integrations.

Release-please will open a new PR whenever there is a releaseable change and it updates the CHANGELOG & other files automatically in this PR. Once the PR is merged a new tag is created.

I have already configured the required `HCLOUD_BOT_TOKEN` secret.


jooola (approved):
Are there links to the changelog that needs to be updated ?

jooola:
> Unfortunately we have a failing CI because the # x-release-please-version comment is being stripped by the kustomize build command, causing a diff between the commit files & the generated ones.

Could a block solve the problem, or does it strip all comments ?
https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files

apricote:
> > Unfortunately we have a failing CI because the # x-release-please-version comment is being stripped by the kustomize build command, causing a diff between the commit files & the generated ones.
> 
> Could a block comment solve the problem, or does it strip all comments ? https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files

Strips all comments. I plan on adding the comment back to the two affected lines before running the diff. This will change anyway once we take care of the helm chart #369

PR #491, opened 2023-09-12
feat: test against Kubernetes 1.28 and drop 1.24

apricote:
In line with our [Kubernetes Versioning Policy](https://github.com/hetznercloud/csi-driver/tree/main/docs/kubernetes#versioning-policy).

### DIFF
--- a/test/e2e/kubernetes/testdriver-1.25.yaml
+++ b/test/e2e/kubernetes/testdriver-1.25.yaml
@@ -8,7 +8,7 @@ DriverInfo:
   Capabilities:
     # Available capabilities are defined in the kubernetes repositories, make
     # sure to select the corresponding k8s version:
-    # https://github.com/kubernetes/kubernetes/blob/release-1.23/test/e2e/storage/framework/testdriver.go#L150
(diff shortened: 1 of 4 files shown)


test/e2e/kubernetes/testdriver-1.25.yaml
```diff
@@ -21,6 +21,7 @@ DriverInfo:
     controllerExpansion: true
     nodeExpansion: true
     onlineExpansion: false
+    offlineExpansion: true
```
jooola, line 24: Is this required by the upgrade?

jooola (approved):

test/e2e/kubernetes/testdriver-1.25.yaml
apricote: This file list available capabilities of the csi-driver that the kubernetes csi test suite will verify. We can only specify capabilities that are added in the lowest-tested version (at least with the current setup). This PR moves this from 1.23 to 1.25. The only capability added is `offlineExpansion`, but this already defaulted to `true`, so everything should keep working as expected.

PR #492, opened 2023-09-12
ci: fix flake in dev/e2e env where subnet is not available

apricote:
The e2e tests fail regularly with [error message](https://github.com/hetznercloud/csi-driver/actions/runs/6155600692/job/16702755191?pr=491#step:7:63):

    hcloud: network 3334615 has no free IP available or is in a different network zone (invalid_input)

As far as I can tell, this happens because the network does not yet have an available subnet when the server is created. With this change, we will wait until the network has the subnet (by grepping for the subnet cidr) before creating the server.


apricote:
Once the CI is successful and this is approved, we should also apply this change in https://github.com/hetznercloud/hcloud-cloud-controller-manager.

jooola (approved):

PR #501, opened 2023-09-14
refactor: deprecated github.com/go-kit/kit/log import

jooola:
The github.com/go-kit/kit/log import is deprecated and recommends using github.com/go-kit/log instead.

See https://pkg.go.dev/github.com/go-kit/kit/log#section-readme


jooola:
Hmm, the `id-token` permission is properly set, I didn't expect TPS to fail for me.

apricote:
> Hmm, the id-token permission is properly set, I didn't expect TPS to fail for me. 

Yea, it does not matter who is the owner, as long as its not from inside the repo the secrets wont be set.

apricote (commented):
Nice, this is missing an update to the `go.mod` to replace the entry for `go-kit/kit`, I will push it to save you some time :)

Issue #473, opened 2023-08-06
Documentation missing for different kubelet dirs


danielr1996:
When Running a distro with a non standard kubelet dir (e.g. k0s uses `/var/lib/k0s/kubelet`) it is necessary to replace all occurences of `/var/lib/kubelet` with `/var/lib/k0s/kubelet` in the `deployment.yaml`. A hint in the `README.md` would be nice, because the errors messages are not that helpful and took me a good time to figures this out.

apricote:
Related to #357

apricote:
Documentation was added in https://github.com/hetznercloud/csi-driver/pull/536

PR #516, opened 2023-09-30
feat: add node.kubeletDir helm chart config value

asteinba:
This PR adds the config value node.kubeletDir to the helm chart to allow deploying the helm chart onto k8s distributions such as k0s. See e.g. #473


apricote (approved):
Awesome! Thanks for the quick contribution @asteinba :)

I have pushed another commit that adds the k0s value to the "full" Snapshot, so we can verify that changing the value actually has the expected effect.

Don't worry about the failing e2e tests, they are currently broken for PRs from forks, but nothing in here should fail the e2e tests and they will run again on `main`.

PR #517, opened 2023-09-30
docs: fix chart name

morremeyer:
The chart's name is not correct in the install documentation, this change sets the correct name


apricote (approved):

apricote:
Thanks a lot @morremeyer!

morremeyer:
@apricote Thanks for the chart, it enables me to deprecate the one from community-tooling!

Issue #534, opened 2023-11-07
Missing "mount" directory on provisioned persistent volume


2martens:
I installed a MicroK8s cluster on a Hetzner server and integrated the CSI driver helm chart. The PV creation itself works and it is mounted to the server by Hetzner. But the directory contains only `vol_data.json` and not a subdirectory `mount` which is used by Kubernetes as mount point for persistent volumes.

I had to manually create this directory via SSH access on the host and then the pod was able to actually mount the PV into itself. The PV was already bound to the PVC. 

PVC created following a volume claim template from Prometheus Operator:
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: >-
    prometheus-prometheus-kube-prometheus-prometheus-db-prometheus-prometheus-kube-prometheus-prometheus-0
  namespace: observability
  uid: dbb1b4b6-0df5-45eb-90e6-ca6d4ae3e767
  resourceVersion: '279650'
  creationTimestamp: '2023-11-06T21:56:13Z'
  labels:
    app.kubernetes.io/instance: prometheus-kube-prometheus-prometheus
    app.kubernetes.io/managed-by: prometheus-operator
    app.kubernetes.io/name: prometheus
    operator.prometheus.io/name: prometheus-kube-prometheus-prometheus
    operator.prometheus.io/shard: '0'
    prometheus: prometheus-kube-prometheus-prometheus
  annotations:
    pv.kubernetes.io/bind-completed: 'yes'
    pv.kubernetes.io/bound-by-controller: 'yes'
    volume.beta.kubernetes.io/storage-provisioner: csi.hetzner.cloud
    volume.kubernetes.io/selected-node: k8s-test-node-0
    volume.kubernetes.io/storage-provisioner: csi.hetzner.cloud
  finalizers:
    - kubernetes.io/pvc-protection
  managedFields:
    - manager: kubelite
      operation: Update
      apiVersion: v1
      time: '2023-11-06T21:56:18Z'
      fieldsType: FieldsV1
      fieldsV1:
        f:metadata:
          f:annotations:
            .: {}
            f:pv.kubernetes.io/bind-completed: {}
            f:pv.kubernetes.io/bound-by-controller: {}
            f:volume.beta.kubernetes.io/storage-provisioner: {}
            f:volume.kubernetes.io/selected-node: {}
            f:volume.kubernetes.io/storage-provisioner: {}
          f:labels:
            .: {}
[...]

PR #536, opened 2023-11-08
docs(k8s): non-standard kubelet dir (k0s, microk8s)

apricote:
Add some explanation about the kubelet dir value so users have an easier time figuring this out.

Related to #534

### DIFF
--- a/docs/kubernetes/README.md
+++ b/docs/kubernetes/README.md
@@ -69,6 +69,16 @@
    kubectl exec -it my-csi-app -- /bin/sh
    ```
 
+### Alternative Kubelet Directory
+
+Some Kubernetes distributions use a non-standard path for the Kubelet directory.
+The csi-driver needs to know about this to successfully mount volumes. You can
+configure this through the Helm Chart Value `node.kubeletDir`.
+
+- Standard: `/var/lib/kubelet`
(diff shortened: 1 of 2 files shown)


docs/kubernetes/README.md
```diff
@@ -69,6 +69,16 @@
...
 
+### Alternative Kubelet Directory
+
+Some Kubernetes distributions use a non-standard path for the Kubelet directory.
+The csi-driver needs to know about this to successfully mount volumes. You can
+configure this through the Helm Chart Value `node.kubeletDir`.
+
+- Standard: `/var/lib/kubelet`
+- **k0s**: `/var/lib/k0s/kubelet`
+- **microk8s**: `/var/snap/microk8s/common/var/lib/kubelet`
```
jooola, line 80: Is microk8s always distributed through snap ? Do we have links to the upstream docs ?

apricote: As far as I can tell microk8s is only distributed through Snap on Linux (its a Ubuntu product afterall): https://microk8s.io/docs/getting-started

Their official documentation for the NFS CSI driver also uses this path: https://microk8s.io/docs/how-to-nfs#install-the-csi-driver-for-nfs-2

jooola (approved):

PR #550, opened 2023-12-17
docs: add instructions for LUKS encryption on Nomad

zefixlluja:
Explains how to enable LUKS encryption for hcloud volumes.

### DIFF
--- a/docs/nomad/README.md
+++ b/docs/nomad/README.md
@@ -285,3 +285,19 @@ job "mariadb" {
 ```sh
 nomad job run mariadb.hcl
 ```
+
+### Volumes Encrypted with LUKS
+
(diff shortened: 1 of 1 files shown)


docs/nomad/README.md
```diff
@@ -285,3 +285,19 @@ job "mariadb" {
 ```sh
 nomad job run mariadb.hcl
 ```
+
+### Volumes Encrypted with LUKS
```
jooola, line 289:
```suggestion
### Volumes encryption with LUKS
```

jooola (changes requested):
Thanks for the contribution, only a small nit and I think we are good to merge this.

jooola (approved):

PR #562, opened 2024-01-12
ci: run tests on main to get coverage comparison

apricote:
Tests are currently only executed for PRs. This has two downsides:

- Coverage comparison in PRs is not available
- If two PRs modify similar code, they have the potential to break tests, this would only be discovered once a new PR is opened.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,10 @@
 name: Run tests
-on: [ pull_request ]
+
+on:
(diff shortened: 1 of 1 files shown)


jooola:
Haha, this was on my radar, thanks for fixing this.

jooola (approved):

apricote:
Hehe, stumbled upon this while checking out the Codecov dashboard.

PR #600, opened 2024-04-30
refactor: use new client.Action.WaitFor

jooola:
This use the new `client.Action.WaitFor` API for waiting actions.


apricote (approved):

jooola:
Part of #605

PR #616, opened 2024-05-28
feat(chart): separate image name and tag

simonostendorf:
This change allows the user to configure the image name separately from the image tag.

This is useful if you only want to change the image path (e.g. private image proxy) and not the tag.

The use of `.tag` is optional and will only be added is a value is set to maintain backward compatibility.

### DIFF
--- a/chart/templates/_common_images.tpl
+++ b/chart/templates/_common_images.tpl
@@ -3,7 +3,7 @@ Return the proper Container Image Name
 {{ include "common.images.image" (dict "value" .Values.controller.image.hcloudCSIDriver "context" $) }}
 */}}
 {{- define "common.images.image" -}}
-{{ .value.name }}{{ if .value.tag }}:{{ tpl .value.tag .context }}{{ end }}
+{{ tpl .value.name .context }}{{ if .value.tag }}:{{ tpl .value.tag .context }}{{ end }}
 {{- end -}}
 
 {{/*
--- a/chart/templates/_common_images.tpl
+++ b/chart/templates/_common_images.tpl
@@ -1,5 +1,5 @@
 {{/*
-Return the proper Container Image Name
+Return the Container Image Name
 {{ include "common.images.image" (dict "value" .Values.controller.image.hcloudCSIDriver "context" .) }}
 */}}
 {{- define "common.images.image" -}}
--- a/chart/templates/_common_images.tpl
+++ b/chart/templates/_common_images.tpl
@@ -1,4 +1,12 @@
 {{/*
+Return the proper Container Image Name
+{{ include "common.images.image" (dict "value" .Values.controller.image.hcloudCSIDriver "context" $) }}
+*/}}
+{{- define "common.images.image" -}}
+{{ .value.name }}{{ if .value.tag }}:{{ tpl .value.tag .context }}{{ end }}
+{{- end -}}
+
+{{/*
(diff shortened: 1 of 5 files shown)


chart/templates/controller/deployment.yaml
```diff
@@ -73,7 +73,7 @@ spec:
         {{- end }}
       containers:
         - name: csi-attacher
-          image: {{ tpl .Values.controller.image.csiAttacher.name . }}
+          image: {{ tpl .Values.controller.image.csiAttacher.name . | default (printf "%s:%s" .Values.controller.image.csiAttacher.repository .Values.controller.image.csiAttacher.tag) }}
```
apricote, line 76: Could you make this a function in `chart/_common_images.tpl` so it can be reused across all images?

apricote (changes requested):

chart/values.yaml
```diff
@@ -71,7 +83,9 @@ controller:
       pullSecrets: []
 
     csiResizer:
-      name: registry.k8s.io/sig-storage/csi-resizer:v1.7.0
+      name: ""
+      repository: registry.k8s.io/sig-storage/csi-resizer
+      tag: v1.7.0
```
jooola, line 88: Do you think it makes sens to reuse the `name` value, and only append the tag, if a `tag` value is present? This way we don't use the `repository` value, that I find a bit confusing.

simonostendorf: Yes, reusing name for the registry + repository sounds like a good idea. Then this PR will only separate the tag from the image name.

chart/templates/controller/deployment.yaml
simonostendorf: I think this is no longer relevant after applying @jooola's comment.

If you think anything could be moved to a template function, please unresolve this conversation and let me know.

chart/templates/node/daemonset.yaml
```diff
@@ -74,7 +74,7 @@ spec:
         {{- end }}
       containers:
         - name: csi-node-driver-registrar
-          image: {{ tpl .Values.node.image.csiNodeDriverRegistrar.name . }}
+          image: {{ tpl .Values.node.image.csiNodeDriverRegistrar.name . }}:{{ .Values.node.image.csiNodeDriverRegistrar.tag }}
```
jooola, line 77: Sorry I didn't express myself correctly. I think this is not backward compatible, we should check if the tag is set, and only then append `:{{ tag }}`. Right now we might end up with `{{ name }}:v2.9.0:`.

jooola: I am going to push the required fix.

apricote (approved):

chart/templates/_common_images.tpl
```diff
@@ -1,3 +1,11 @@
+{{/*
+Return the proper Container Image Name
+{{ include "common.images.image" (dict "value" .Values.controller.image.hcloudCSIDriver "context" $) }}
```
jooola, line 3:
```suggestion
{{ include "common.images.image" (dict "value" .Values.controller.image.hcloudCSIDriver "context" .) }}
```

chart/templates/_common_images.tpl
```diff
@@ -1,3 +1,11 @@
+{{/*
+Return the proper Container Image Name
```
jooola, line 2:
```suggestion
Return the Container Image Name
```

PR #623, opened 2024-06-13
fix(resize): recover from part-way failed resizes

apricote:
If a previous resize attempt failed after calling `POST /v1/volumes/{id}/resize` (rate limit, network connectivity, ...), the volume might already have the target size.

The PVC (in Kubernetes) would still have the old size in its status, so the `external-resizer` makes another attempt at the resize. In the Hetzner Cloud API, a resize must always be larger than the current size, so the new resize failed with:

    volume size is too small (invalid_input)

This is now fixed, as we compare the size of the Hetzner Cloud volume to the desired size and only proceed if an actual resize is needed.

This was a violation of the CSI Spec[0], which stated for `ControllerExpandVolume`:

> This operation MUST be idempotent. If a volume corresponding to the specified volume ID is already larger than or equal to the target capacity of the expansion request, the plugin SHOULD reply 0 OK.

[0] https://github.com/container-storage-interface/spec/blob/release-1.9/spec.md#controllerexpandvolume

### DIFF
--- a/api/volume.go
+++ b/api/volume.go
@@ -356,42 +356,29 @@ func (s *VolumeService) Resize(ctx context.Context, volume *csi.Volume, size int
 
 	hcloudVolume, _, err := s.client.Volume.GetByID(ctx, volume.ID)
 	if err != nil {
-		level.Info(logger).Log(
-			"msg", "failed to get volume",
-			"err", err,
-		)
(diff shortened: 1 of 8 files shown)


jooola (approved):

apricote:
Approved from my side too, as soon as the e2e tests are green.

PR #628, opened 2024-06-20
test: implement new development environment

jooola:
Replace the multiple bash scripts used to setup the `csi-driver` test environment with Terraform and Make


.github/workflows/test.yml
```diff
@@ -40,53 +40,72 @@ jobs:
...
-
-          trap "hack/dev-down.sh" EXIT
-          source <(hack/dev-up.sh)
+          source env.sh
 
-          skaffold build --tag="e2e-${GITHUB_RUN_ID}-${GITHUB_RUN_NUMBER}"
-          tag=$(skaffold build --tag="e2e-${GITHUB_RUN_ID}-${GITHUB_RUN_NUMBER}" --quiet --output="{{ (index .Builds 0).Tag }}")
+          skaffold build --tag="$TAG"
+          tag=$(skaffold build --tag="$TAG" --quiet --output="{{ (index .Builds 0).Tag }}")
           skaffold deploy --images=docker.io/hetznercloud/hcloud-csi-driver=$tag
```
apricote, line 101: We can replace the three skaffold command with a simple `skaffold run`

dev/hcloud-k8s-env/main-setup.tf
```diff
@@ -0,0 +1,216 @@
...
+          --disable=traefik \
+          --flannel-backend=none \
+          --node-external-ip='${hcloud_server.control.ipv4_address}' \
+          --node-ip='${hcloud_server_network.control.ip}'" \
+        --local-path='${local.kubeconfig_path}'
+    EOT
+  }
+}
+
+resource "null_resource" "k3sup_worker" {
```
apricote, line 62: Was the `depends_on` on Cilium+HCCM Helm Charts no longer necessary?

apricote (approved):

dev/hcloud-k8s-env/main-setup.tf
jooola: Yes, I tested without it, and the whole problem was related to the nodes hostname.

PR #641, opened 2024-06-24
ci: cleanup resources after e2e tests

apricote:
Remove all deployed resources to avoid running into limit errors.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,6 +102,11 @@ jobs:
           source dev/files/env.sh
           make -C test/e2e/kubernetes serial
 
+      - name: Cleanup
+        if: ${{ always() }}
+        continue-on-error: true
+        run: make -C dev down
+
   deploy-manifests:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -102,6 +102,11 @@ jobs:
           source dev/files/env.sh
           make -C test/e2e/kubernetes serial
 
+      - name: Cleanup
+        if: ${{ always() }}
```
jooola, line 106:
```suggestion
        if: always()
```

jooola (approved):

jooola:
Hmm, does the k8s tests cleanup the volumes if the test fail? Because I am not sure if we clean volumes with the terraform module.

apricote:
We dont, the kubernetes csi tests try to clean those up if possible. I could add a `hcloud volume delete $(hcloud volume list -o noheader -o columns=id)` after taking down the env. (With a check for empty list)

jooola:
> We dont, the kubernetes csi tests try to clean those up if possible. I could add a hcloud volume delete $(hcloud volume list -o noheader -o columns=id) after taking down the env. (With a check for empty list)

Let's see how that goes, if we stumble on a problem again we'll add the command.

PR #647, opened 2024-06-27
chore: reuse terraform k8s-dev module

jooola:
Load the k8s-dev module from https://github.com/hetznercloud/terraform-k8s-dev.

Copy over the example with latest best practices.


dev/main.tf
```diff
@@ -1,5 +1,5 @@
 module "dev" {
-  source = "./hcloud-k8s-env"
+  source = "github.com/hetznercloud/terraform-k8s-dev?ref=v0.2.0"
```
apricote, line 2: Should we add renovate here? I think we need a new manager for `.tf` files

apricote (approved):

dev/main.tf
apricote: Nevermind, this should be supported: https://docs.renovatebot.com/modules/manager/terraform/#modules

PR #660, opened 2024-07-12
test: add integration test tag

jooola:
The integration test need docker to run, we use a tag to filter out integrations tests in environment that do not have docker.


apricote (changes requested):
Could you update the README with this tag? https://github.com/hetznercloud/csi-driver?tab=readme-ov-file#run-the-docker-e2e-tests

apricote (approved):

PR #673, opened 2024-07-25
feat: truncate poll exponential backoff function to max 5s

jooola:
Related to #346 and #380

The initial exponential back off algorithm introduced in #380 was not truncated, which could lead to very long times quickly. The recent release of hcloud-go truncated the default exponential back off algorithm to 60s: https://github.com/hetznercloud/hcloud-go/commit/fd1f46cc35e61dde1e524399eef88c38a757636e

If we take the scenario described in #346, I think we can still reduce the max interval value for the exponential back off algorithm to <30s.


jooola:
Happy to increase the max to something between 5s and 30s

Issue #400, opened 2023-03-27
No topology key found on hw nodes


SNB-hz:
In clusters with hardware nodes, a new PVC and its workload can be stuck in `Pending` state if they are scheduled without nodeAffinity.

**Steps to reproduce:**
- run a cluster that includes a hardware worker, and label the hw node with `instance.hetzner.cloud/is-root-server=true` as mentioned in the README 
- install CSI driver according to instructions
- apply the test-pvc and pod mentioned in the README, using the default storageClass with `WaitForFirstConsumer` volumeBindingMode

**Expected Behaviour:**

hcloud-csi-controller should provide the desired / required topology constaints to the k8s scheduler, which then schedules the pod on a node fulfilling the topology requirements.
As the hardware node does not run csi-driver and cannot mount hetzner cloud volumes, the workload should not be scheduled there.

**Observed Behaviour:**

- Both pvc and pod are stuck in `Pending` state.
- the container `csi-provisioner` of the CSI Controller deployment logs this Error:
```
'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "hcloud-volumes": error generating accessibility requirements: no topology key found on CSINode hardwarenode.testcluster
```

**More Info:**
Tested with csi-driver  2.1.1 as well as 2.2.0, together with csi-provisioner 3.4.0

- the DaemonSet for `hcloud-csi-node` does not run on the hw node
- because of this, the `csinode` object for the node lists no driver:
```
kubectl get csinode
NAME                     DRIVERS       AGE
virtualnode.testcluster     1           1d
hardwarenode.testcluster    0           1d
```
- the `csinode` object of the virtual node looks ok:
```
kubectl get csinode virtualnode.testcluster -oyaml
apiVersion: storage.k8s.io/v1
kind: CSINode
...
spec:
  drivers:
  - allocatable:
      count: 16
    name: csi.hetzner.cloud
    nodeID: "12769030"
    topologyKeys:
    - csi.hetzner.cloud/location
```
- the `csinode` object of the hardware node does not have a driver and therefore no topology key, as the node intentionally runs no `hcloud-csi-node` pod due to the `nodeAffinity`:

```
kubectl get csinode hardwarenode.testcluster -oyaml
apiVersion: storage.k8s.io/v1
kind: CSINode
...
spec:
  drivers: null
```

**Theory**

It seems we are hitting [this Issue in csi-provisioner](https://github.com/kubernetes-csi/external-provisioner/issues/544).
[...]

PR #743, opened 2024-10-09
feat: force pods with volumes to be scheduled on Cloud servers

lukasmetzner:
Due to a bug in the scheduler a node with no driver instance might be picked and the volume is stuck in pending as the "no capacity - > reschedule" recovery is never triggered [[0]](https://github.com/kubernetes/kubernetes/pull/122109), [[1]](https://github.com/kubernetes-csi/external-provisioner/issues/544).

- See #400

### DIFF
--- a/chart/.snapshots/full.values.yaml
+++ b/chart/.snapshots/full.values.yaml
@@ -371,3 +371,4 @@ storageClasses:
     defaultStorageClass: false
     reclaimPolicy: Keep
     allowedTopologyCloudServer: false
+
--- a/chart/.snapshots/full.values.yaml
+++ b/chart/.snapshots/full.values.yaml
@@ -370,3 +370,4 @@ storageClasses:
   - name: foobar
     defaultStorageClass: false
     reclaimPolicy: Keep
+    allowedTopologyCloudServer: false
--- a/chart/templates/core/storageclass.yaml
+++ b/chart/templates/core/storageclass.yaml
@@ -16,6 +16,10 @@ allowedTopologies:
   - key: instance.hetzner.cloud/is-root-server
     values:
     - "false"
+- matchLabelExpressions:
+  - key: instance.hetzner.cloud/provided-by
+    values:
+    - "cloud"
 {{- end }}
 ---
 {{- end }}
--- a/chart/templates/core/storageclass.yaml
+++ b/chart/templates/core/storageclass.yaml
@@ -10,6 +10,13 @@ provisioner: csi.hetzner.cloud
 volumeBindingMode: WaitForFirstConsumer
 allowVolumeExpansion: true
 reclaimPolicy: {{ $val.reclaimPolicy | quote }}
+{{- if $val.includesRobotServers }}
+allowedTopologies:
+- matchLabelExpressions:
+  - key: instance.hetzner.cloud/is-root-server
+    values:
+    - "false"
+{{- end }}
 ---
 {{- end }}
 {{- end }}
--- a/chart/templates/core/storageclass.yaml
+++ b/chart/templates/core/storageclass.yaml
@@ -10,7 +10,7 @@ provisioner: csi.hetzner.cloud
 volumeBindingMode: WaitForFirstConsumer
 allowVolumeExpansion: true
 reclaimPolicy: {{ $val.reclaimPolicy | quote }}
-{{- if $val.includesRobotServers }}
+{{- if $val.allowedTopologyCloudServer }}
 allowedTopologies:
 - matchLabelExpressions:
   - key: instance.hetzner.cloud/is-root-server
--- a/chart/templates/core/storageclass.yaml
+++ b/chart/templates/core/storageclass.yaml
@@ -13,10 +13,6 @@ reclaimPolicy: {{ $val.reclaimPolicy | quote }}
 {{- if $val.allowedTopologyCloudServer }}
 allowedTopologies:
 - matchLabelExpressions:
-  - key: instance.hetzner.cloud/is-root-server
-    values:
-    - "false"
-- matchLabelExpressions:
   - key: instance.hetzner.cloud/provided-by
     values:
     - "cloud"
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -728,4 +728,5 @@ storageClasses:
   - name: hcloud-volumes
     defaultStorageClass: true
     reclaimPolicy: Delete
-    includesRobotServers: false 
+    ## @param storageClass.allowedTopologyCloudServer Prevents pods from being scheduled on nodes, specifically Robot servers, where Hetzner volumes are unavailable
+    allowedTopologyCloudServer: false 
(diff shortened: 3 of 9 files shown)


docs/kubernetes/README.md
```diff
@@ -189,8 +189,26 @@ $ kubectl apply -f https://raw.githubusercontent.com/hetznercloud/csi-driver/v2.
...
+    defaultStorageClass: true
+    reclaimPolicy: Delete
+    includesRobotServers: true # <---
+```
+
+Label the nodes according to their type:
+
+**Cloud Servers**
+```bash
+kubectl label nodes <node name> instance.hetzner.cloud/is-root-server=false
```
apricote, line 208: Considering that we have users with clusters with even more mixed servers (Cloud, Robot, Raspberry Pis at home, ...) I think this nomenclature is confusing.

What do you think about adding a new annotation `instance.hetzner.cloud/{source,type,service,owner}: {cloud,robot}` that we can set in HCCM and use here?

docs/kubernetes/README.md
```diff
@@ -189,8 +189,43 @@ $ kubectl apply -f https://raw.githubusercontent.com/hetznercloud/csi-driver/v2.
...
+
+```yaml
+storageClasses:
+  - name: hcloud-volumes
+    defaultStorageClass: true
+    reclaimPolicy: Delete
+    includesRobotServers: true # <---
+```
+
+Label the nodes according to their type:
```
apricote, line 204: Should we also add a hint here a la "If you use hcloud-cloud-controller-manager v1.20.0+ these labels are added automatically"?

chart/templates/core/storageclass.yaml
```diff
@@ -10,6 +10,17 @@ provisioner: csi.hetzner.cloud
 volumeBindingMode: WaitForFirstConsumer
 allowVolumeExpansion: true
 reclaimPolicy: {{ $val.reclaimPolicy | quote }}
+{{- if $val.includesRobotServers }}
+allowedTopologies:
+- matchLabelExpressions:
+  - key: instance.hetzner.cloud/is-root-server
+    values:
+    - "false"
```
apricote, line 18: With the new labels, I would just remove `is-root-server=false`

chart/templates/core/storageclass.yaml
```diff
@@ -10,6 +10,17 @@ provisioner: csi.hetzner.cloud
 volumeBindingMode: WaitForFirstConsumer
 allowVolumeExpansion: true
 reclaimPolicy: {{ $val.reclaimPolicy | quote }}
+{{- if $val.includesRobotServers }}
+allowedTopologies:
```
apricote, line 14:
```suggestion
{{- if $val.allowedTopologyCloudServer }}
allowedTopologies:
```

I find `includesRobotServers` very confusing. Setting this to `true` **excludes** Robot servers from being allowed the storage class.

chart/values.yaml
```diff
@@ -724,3 +728,4 @@ storageClasses:
   - name: hcloud-volumes
     defaultStorageClass: true
     reclaimPolicy: Delete
+    includesRobotServers: false
```
apricote, line 731: Can you add some documentation to this value to explain what it does and that it is immutable (ie. can only be set when creating a storage class, not modified later on).

chart/templates/core/storageclass.yaml
lukasmetzner: Should we also remove the deprecation notice from the docs?
https://github.com/hetznercloud/csi-driver/blob/052842a3f39341e7db4413da30c7a5383369593f/docs/kubernetes/README.md?plain=1#L219-L231

chart/values.yaml
lukasmetzner: Added parameter documentation to the helm values and the readme.md.

docs/kubernetes/README.md
```diff
@@ -189,8 +189,43 @@ $ kubectl apply -f https://raw.githubusercontent.com/hetznercloud/csi-driver/v2.
 
 ## Integration with Root Servers
 
-Root servers can be part of the cluster, but the CSI plugin doesn't work there. Taint the root server as follows to skip that node for the DaemonSet.
+Root servers can be part of the cluster, but the CSI plugin doesn't work there and the current behaviour of the scheduler can cause Pods to be stuck in `Pending`. 
 
+Setting `allowedTopologyCloudServer` to true prevents pods from being scheduled on nodes, specifically Robot servers, where Hetzner volumes are unavailable. This value can not be changed after the initial creation of a storage class.
```
apricote, line 194: I would highlight that this option is from the Helm Chart.

```suggestion
In the Helm Chart you can set `allowedTopologyCloudServer` to true to prevent pods from being scheduled on nodes, specifically Robot servers, where Hetzner volumes are unavailable. This value can not be changed after the initial creation of a storage class.
```

docs/kubernetes/README.md
```diff
@@ -189,8 +189,43 @@ $ kubectl apply -f https://raw.githubusercontent.com/hetznercloud/csi-driver/v2.
...
+storageClasses:
+  - name: hcloud-volumes
+    defaultStorageClass: true
+    reclaimPolicy: Delete
+    allowedTopologyCloudServer: true # <---
+```
+
+To ensure proper topology evaluation, labels are needed to indicate whether a node is a cloud VM or a dedicated server from Robot. If you are using the `hcloud-cloud-controller-manager` version 1.20.0 or later, these labels are added automatically. Otherwise, you will need to label the nodes manually.
+
+### New Label
```
apricote, line 206:
```suggestion
### Adding labels manually
```

chart/.snapshots/full.values.yaml
```diff
@@ -370,3 +370,4 @@ storageClasses:
   - name: foobar
     defaultStorageClass: false
     reclaimPolicy: Keep
+    allowedTopologyCloudServer: false
```
apricote, line 373: Could you add an empty new line here?

docs/kubernetes/README.md
```diff
@@ -189,8 +189,43 @@ $ kubectl apply -f https://raw.githubusercontent.com/hetznercloud/csi-driver/v2.
...
+
+```yaml
+storageClasses:
+  - name: hcloud-volumes
+    defaultStorageClass: true
+    reclaimPolicy: Delete
+    allowedTopologyCloudServer: true # <---
+```
+
+To ensure proper topology evaluation, labels are needed to indicate whether a node is a cloud VM or a dedicated server from Robot. If you are using the `hcloud-cloud-controller-manager` version 1.20.0 or later, these labels are added automatically. Otherwise, you will need to label the nodes manually.
```
apricote, line 204: Next HCCM version is 1.21.0: https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/723

```suggestion
To ensure proper topology evaluation, labels are needed to indicate whether a node is a cloud VM or a dedicated server from Robot. If you are using the `hcloud-cloud-controller-manager` version 1.21.0 or later, these labels are added automatically. Otherwise, you will need to label the nodes manually.
```

chart/templates/core/storageclass.yaml
apricote: Makes sense to at least state that we now prefer the new label, even though there is no immediate reason to remove the old label.

Issue #382, opened 2023-02-17
Docker Swarm support: Remove mock staging/unstaging once possible


s4ke:
In the original PR #376 we added a workaround where we added mock staging/unstaging support that will not be needed anymore once https://github.com/moby/swarmkit/pull/3116 is merged.

This issue is here to track this TODO.

s4ke:
This should work without by

apricote:
Hey @s4ke, could you implement this?

s4ke:
Yes I can, but I have to see where it fits my schedule. Any hard deadline for when you want this in?

apricote:
Wonderful :) No deadline, would just love to see this done before the stale bot tries to close it again (~3 months)

s4ke:
May I ask you to remind me if this is still not done in let's say, 1 month?

apricote:
Well, I forgot about it :D @s4ke ping.

s4ke:
Same here. If anyone else wants to take this, this is probably also a good first issue since it is just about removing operations and testing once against a Swarm.

apricote:
If anyone wants to send a PR:

- Revert commit https://github.com/hetznercloud/csi-driver/pull/376/commits/619fa5c3996e17e84b09deca0c956229e5554c77
- Test against a current version of Docker Swarm

PR #376, opened 2023-02-07
add Docker Swarm support

s4ke:
ready for review

closes https://github.com/hetznercloud/csi-driver/issues/374

PR #746, opened 2024-10-10
feat(swarm): removed workaround support for mock staging/unstaging

lukasmetzner:
With the merge of https://github.com/moby/swarmkit/pull/3116 the mock staging/unstaging is not needed anymore. It was introduced in our csi-driver in this commit: https://github.com/hetznercloud/csi-driver/commit/619fa5c3996e17e84b09deca0c956229e5554c77. Which is part of a large PR (squashed commit) that included experimental support for Docker swarm.

See:
- #376 
- #382

### DIFF
--- a/deploy/docker-swarm/pkg/config.json
+++ b/deploy/docker-swarm/pkg/config.json
@@ -26,12 +26,6 @@
         "value"
       ],
       "value": "debug"
-    },
-    {
-      "name": "FORCE_STAGING_SUPPORT",
-      "description": "workaround: force staging support to make Docker 23.0.0 work without https://github.com/moby/swarmkit/pull/3116",
-      "settable": ["value"],
-      "value": "true"
     }
   ],
   "interface": {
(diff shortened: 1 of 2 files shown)


jooola:
Is this change backward compatible for previous docker versions?

Do we state somewhere that the docker swarm support is experimental? I could only find a statement that we do not provide official support for it.

jooola:
Changed the scope to feat as we have some user facing changes.

apricote:
> Is this change backward compatible for previous docker versions?

This is only compatible with docker-swarm versions that have the fix from https://github.com/moby/swarmkit/pull/3116. Not sure which versions of docker-swarm include that.


> Do we state somewhere that the docker swarm support is experimental? I could only find a statement that we do not provide official support for it.

In the docker swarm docs it is mentioned that the support is in **beta**, though I would say that "beta" is not defined for us. We do not test it and do not publish a version compatibility for docker swarm.

apricote (approved):
Nice cleanup :)

PR #747, opened 2024-10-10
feat: allow passing mkfs format options via storage class parameters

lukasmetzner:
User can now pass format options directly to mkfs via storage class parameters for every fstype.

Older kernels might not be compatible with the mkfs.xfs options we set. As mkfs.xfs provides a default set of options for certain linux kernel lts versions we can choose the appropriate set of values by inspecting the current kernel version.

If we can find a suitable config we fall back to the most compatible version with older kernels.

BEGIN_COMMIT_OVERRIDE
feat: allow passing mkfs format options via storage class parameters (#747)
feat: change XFS default options to support older kernels (#747)
END_COMMIT_OVERRIDE


internal/volumes/mount.go
```diff
@@ -203,10 +203,19 @@ func (s *LinuxMountService) FormatDisk(disk string, fstype string) error {
...
+			_, _, err = command(
+				"mkfs.xfs", "-f", "-c", fmt.Sprintf("options=%s", xfsConfigPath), disk)
+		} else {
+			// Fallback
+			// Default flags extracted from /usr/share/xfsprogs/mkfs/lts_4.19.conf
+			// for maximum backwards compatibility with older kernels
+			// Also see: https://man.archlinux.org/man/xfs_admin.8.en#O
+			// This document mentions the flags and their minimal supported kernel version
+			_, _, err = command(
+				"mkfs.xfs", "-i", "nrext64=0", "-m", "bigtime=0", "-m", "inobtcount=0", disk)
```
jooola, line 217: Should we allow the users to set a cluster wide supported kernel version?

For example, if a node has a specific kernel version, creates and formats the volume, but then the volumes then moves to a different node with an older kernel version. Maybe having a cluster wide version can be handy for those use cases?

lukasmetzner:
Memo: identify config once on node start. Probably no need to do this on every formatdisk call.

internal/volumes/mount.go
lukasmetzner: So we:
1. Use the cluster wide configured kernel version
2. Use the current nodes kernel version as hopefully he has uniform cluster

internal/volumes/utils.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func GetXFSConfigPath(current *KernelVersion) string {
+	files, err := os.ReadDir(XFSDefaultConfigsLocation)
+	if err != nil {
+		return ""
+	}
+
+	files = slices.DeleteFunc(files, func(file fs.DirEntry) bool {
+		return !strings.HasPrefix(file.Name(), "lts_")
```
jooola, line 88: Could we be more strict about this check? This feels a bit fragile, maybe also checking for the extension is worth it?

jooola, line 88: Maybe you can use https://pkg.go.dev/path/filepath#Glob ?

docs/kubernetes/README.md
```diff
@@ -123,6 +123,38 @@ parameters:
 
 Your nodes might need to have `cryptsetup` installed to mount the volumes with LUKS.
 
+### XFS Filesystem
+
+XFS can be enabled by creating a custom storage class and settting the `csi.storage.k8s.io/fstype` parameter to `xfs`.
+
+If all nodes in your cluster run the same kernel version, the CSI driver will automatically determine the correct `mkfs.xfs` options.
```
jooola, line 130: Should we put this feature behind a flag? The magic of this feature might be too much for some users.

internal/volumes/mount.go
```diff
@@ -203,10 +207,39 @@ func (s *LinuxMountService) FormatDisk(disk string, fstype string) error {
...
+			if err == nil {
+				level.Info(s.logger).Log(
+					"msg", "formatted disk with xfs",
+					"xfs-config-path", xfsConfigPath,
+					"is-cluster-config", xfsMinSupportedKernel != "",
+				)
+			}
+		} else {
+			// Fallback
+			// Default flags extracted from /usr/share/xfsprogs/mkfs/lts_4.19.conf
```
jooola, line 235: I think this should be documented: we only support Linux kernels that are not EOL, so >=4.19 as of today.

We should also mention that this kernel version policy is only related to XFS for now, and maybe that bumping the minimal Linux kernel version will not require a new major version?

internal/volumes/mount.go
jooola: I wonder if we should allow the users to pass/point to their own XFS config file? If we do this, we might simply want to allow users to pass their own `mkfs` arguments, what do you think?

Currently, we rely on the XFS configs from the alpine distribution, I am not sure, but maybe there are some difference between distros that the user want to handle?

Maybe having the kernel version parsing and XFS config magic behind a feature flag is better?

What do you think about:
1. If provided, user xfs config (or mkfs args ?)
2. If provided, user xfs min kernel version 
3. Fallback to the oldest non EOL linux kernel xfs settings

jooola (commented):
I am really unsure what the best way of configuring the xfs settings is.

I wrote a bunch of ideas, let me know what you think. I'd wait for Julian's input before merging this anyways :)

internal/volumes/mount.go
```diff
@@ -203,10 +207,39 @@ func (s *LinuxMountService) FormatDisk(disk string, fstype string) error {
...
+		var err error
+		// By default node checks its own minimal supported kernel version
+		var xfsConfigPath = s.xfsConfigPath
+
+		// User configures cluster wide minimal supported kernel version
+		if xfsMinSupportedKernel != "" {
+			kv, err := ParseKernelVersion(xfsMinSupportedKernel)
+			if err == nil {
+				xfsConfigPath = GetXFSConfigPath(kv)
+			}
```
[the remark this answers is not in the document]
lukasmetzner: The `xfsMinSupportedKernel` parameter which is configured with the storage class is only provided upon a `NodePublishVolumeRequest` in the `VolumeContext`. So we can not really update it on startup. Only thing we can implement on startup is checking the current kernel version and selecting the config based on the information.

internal/volumes/mount.go
lukasmetzner: I agree on 1. and 2.

(3.) For a user who just wants to create a cluster of homogeneous nodes running the same Linux kernel, it may be simpler to rely on the `mkfs.xfs` default arguments instead of opting for the oldest supported configuration. As this way the user gets the latest XFS features.

lukasmetzner: After discussing with @jooola we came with the following possible features:
1. User can pass args to `mkfs.xfs`
2. User can configure cluster wide min kernel version
3. User can set boolean to automatically choose the mkfs args based on the kernel version -> can only be used in homogeneous clusters
4. Default: Use the oldest non EOL linux kernel xfs args

internal/volumes/mount.go
```diff
@@ -134,7 +138,16 @@ func (s *LinuxMountService) Publish(targetPath string, devicePath string, opts M
 		return s.mounter.Mount(devicePath, targetPath, opts.FSType, mountOptions)
 	}
 
-	return s.mounter.FormatAndMount(devicePath, targetPath, opts.FSType, mountOptions)
+	formatOptions := make([]string, 0)
+
+	if opts.FormatOptions != "" {
+		formatOptions = strings.Split(opts.FormatOptions, " ")
```
jooola, line 144: This is not safe, please use a shell lexer to split the arguments, or don't split them at all.

jooola: Or use a list of argument instead of a string as parameter.

dev/main.tf
```diff
@@ -4,6 +4,7 @@ module "dev" {
   name         = "csi-driver-${replace(var.name, "/[^a-zA-Z0-9-_]/", "-")}"
   hcloud_token = var.hcloud_token
   worker_count = 3
+  hcloud_location = "hel1"
```
jooola, line 7:
```suggestion
  hcloud_location = "hel1"
  worker_count = 3
```

docs/kubernetes/README.md
```diff
@@ -123,6 +123,20 @@ parameters:
 
 Your nodes might need to have `cryptsetup` installed to mount the volumes with LUKS.
 
+### Formatting Options
+You can specify extra formatting options which are passed directly to `mkfs.FSTYPE` via the `formatOptions` parameter in the storage class.
```
jooola, line 127:
```suggestion
### Formatting Options

You can specify extra formatting options which are passed directly to `mkfs.FSTYPE` via the `formatOptions` parameter in the storage class.
```

docs/kubernetes/README.md
```diff
@@ -123,6 +123,21 @@ parameters:
...
 
+### Formatting Options
+
+You can specify extra formatting options which are passed directly to `mkfs.FSTYPE` via the `formatOptions` parameter in the storage class.
+
+#### Example
+```yaml
+parameters:
+  csi.storage.k8s.io/fstype: xfs
+  formatOptions: "-i nrext64=1"
```
jooola, line 134: Should this parameter be prefixed with `fs` ? `fsFormatOptions` or even `fsFormatArgs` ?

internal/volumes/mount.go
```diff
@@ -134,7 +139,20 @@ func (s *LinuxMountService) Publish(targetPath string, devicePath string, opts M
...
+	if opts.FormatOptions != "" {
+		lexer := shell.NewLex('\\')
+		formatOptions, err = lexer.ProcessWords(opts.FormatOptions, shell.EnvsFromSlice([]string{}))
+		if err != nil {
+			return err
+		}
+	} else if opts.FSType == "xfs" {
+		formatOptions = append(formatOptions, "-c")
+		formatOptions = append(formatOptions, fmt.Sprintf("options=%s", XFSDefaultConfigPath))
+	}
```
jooola, line 153:
```suggestion
	if opts.FormatOptions != "" {
		lexer := shell.NewLex('\\')
		formatOptions, err = lexer.ProcessWords(opts.FormatOptions, shell.EnvsFromSlice([]string{}))
		if err != nil {
			return err
		}
	}
	if opts.FSType == "xfs" {
		formatOptions = append(formatOptions, "-c")
		formatOptions = append(formatOptions, fmt.Sprintf("options=%s", XFSDefaultConfigPath))
	}
```

Not sure the conditions make sense here?

internal/volumes/mount.go
```diff
@@ -134,7 +139,20 @@ func (s *LinuxMountService) Publish(targetPath string, devicePath string, opts M
...
+
+	if opts.FormatOptions != "" {
+		lexer := shell.NewLex('\\')
+		formatOptions, err = lexer.ProcessWords(opts.FormatOptions, shell.EnvsFromSlice([]string{}))
+		if err != nil {
+			return err
+		}
+	} else if opts.FSType == "xfs" {
+		formatOptions = append(formatOptions, "-c")
+		formatOptions = append(formatOptions, fmt.Sprintf("options=%s", XFSDefaultConfigPath))
```
jooola, line 152:
```suggestion
		formatOptions = append(formatOptions, "-c",  fmt.Sprintf("options=%s", XFSDefaultConfigPath))
```

Append has a variadic elements argument.

internal/volumes/mount.go
lukasmetzner: If the user sets formatOptions when using xfs we probably dont want to append `-c options=/usr/share/xfsprogs/mkfs/lts_4.19.cfg`. So either set format options otherwise use fallback if we format with xfs.

jooola: Makes sens, my bad :)

docs/kubernetes/README.md
lukasmetzner: How about `mkfsFormatOptions`? Then we are very verbose about whats actually happening under the hood. Might also easier for an user to identify what options are supported (going directly to the `mkfs` docs).

.github/workflows/test.yml
```diff
@@ -29,6 +31,12 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
+      # Check if xfs config file exists in docker image
+      - name: Check XFS Config File
+        run: |
+          docker build -t xfs-test:latest -f dev/Dockerfile .
```
apricote, line 37: To really be sure we should test against the production image, but I understand thats harder to do because the binaries are compiled outside of the image. As we also run the e2e tests with the non-prod image it does not make the situation worse than it was before :D

docs/kubernetes/README.md
```diff
@@ -123,6 +123,21 @@ parameters:
...
+#### Example
+```yaml
+parameters:
+  csi.storage.k8s.io/fstype: xfs
+  fsFormatOptions: "-i nrext64=1"
+```
+
+### XFS Filesystem
+
+When using XFS as the filesystem type, we apply a default configuration to mkfs to ensure maximum compatibility with older Linux kernel versions.
```
apricote, line 139: These docs should be a part of the general docs, they are not specific to Kubernetes. Happy to keep them here until we redo the docs though.

internal/volumes/mount.go
```diff
@@ -134,7 +139,19 @@ func (s *LinuxMountService) Publish(targetPath string, devicePath string, opts M
 		return s.mounter.Mount(devicePath, targetPath, opts.FSType, mountOptions)
 	}
 
-	return s.mounter.FormatAndMount(devicePath, targetPath, opts.FSType, mountOptions)
+	formatOptions := make([]string, 0)
+
+	if opts.FsFormatOptions != "" {
+		lexer := shell.NewLex('\\')
+		formatOptions, err = lexer.ProcessWords(opts.FsFormatOptions, shell.EnvsFromSlice([]string{}))
```
apricote, line 146: I would personally find passing list of strings simpler than importing a shell lexer here, not a hard requirement

internal/volumes/mount.go
```diff
@@ -134,7 +139,19 @@ func (s *LinuxMountService) Publish(targetPath string, devicePath string, opts M
...
+		if err != nil {
+			return err
+		}
+	} else if opts.FSType == "xfs" {
+		formatOptions = append(formatOptions, "-c", fmt.Sprintf("options=%s", XFSDefaultConfigPath))
+	}
```
apricote, line 152: To make sure I understood this right: If the user specifies any options at all we pass them along. If they do not specify any options and use `xfs` we set the default config file.

If thats true, this behaviour should be added to the docs:

```
When using XFS as the filesystem type and no `fsFormatOptions` are set, we apply a default configuration to mkfs to ensure maximum compatibility with older Linux kernel versions. This configuration file is from the `xfs-utils` package and currently targets Linux 4.19.

> [!NOTE]
> The targeted minimum Linux Kernel version may be raised in a minor update, we will announce this in the Release Notes.

If you set any options at all, it is your responsible to make sure that all default flags from `mkfs.xfs` are supported on your current Linux Kernel version or that you set the flags appropriately.

```

internal/volumes/mount.go
```diff
@@ -8,11 +8,15 @@ import (
...
 	"github.com/go-kit/log/level"
+	"github.com/moby/buildkit/frontend/dockerfile/shell"
 	"k8s.io/mount-utils"
 	"k8s.io/utils/exec"
 )
 
-const DefaultFSType = "ext4"
+const (
+	DefaultFSType        = "ext4"
+	XFSDefaultConfigPath = "/usr/share/xfsprogs/mkfs/lts_4.19.conf"
```
apricote, line 18:
```suggestion
  // XFSDefaultConfigPath is the oldest Linux Version available from `xfsprogs`. If this becomes unavailable, we need to increase it to the next lowest version and announce the change in the Release Notes.
	XFSDefaultConfigPath = "/usr/share/xfsprogs/mkfs/lts_4.19.conf"
```

internal/volumes/mount.go
lukasmetzner: Unfortunately we can´t do this as StorageClass parameters are unmarshalled into map[string]string.

.github/workflows/test.yml
lukasmetzner: I just realized we can just put two dummy files there. The main point is to test if file is present after installing xfsprogs via the package manager. No need for the actual binaries to be there.

internal/volumes/mount.go
lukasmetzner: Yes this is correct. 

Added to docs :heavy_check_mark:

apricote (approved):

PR #749, opened 2024-10-17
refactor: Switched to k8s.io/mount-utils

lukasmetzner:
Replaced our code for formatting and other disk related functions to k8s.io/mount-utils. This leaves only code that is related to LUKS.

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -114,6 +114,8 @@ func formatDisk(mounter *mount.SafeFormatAndMount, device string, fstype string)
 		return err
 	}
 
+  // The library we use to format volumes only supports a combined "FormatAndMount"
+  // so we call that and immediately unmount and cleanup the temp path afterwards.
 	defer os.RemoveAll(tmppath)
 	defer mounter.Unmount(tmppath)
 	return mounter.FormatAndMount(device, tmppath, fstype, nil)
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -9,6 +9,8 @@ import (
 	"regexp"
 	"strconv"
 	"testing"
+
+	"k8s.io/mount-utils"
 )
 
 const testImageName = "hcloud-csi-driver-integrationtests"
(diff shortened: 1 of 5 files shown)


test/integration/volumes_test.go
```diff
@@ -123,29 +157,39 @@ func TestVolumePublishUnpublish(t *testing.T) {
...
+						if got.Type != expected.Type {
+							t.Fatalf("Expected Mount Error %s, but got %s", expected.Type, got.Type)
+						}
+						return
+					} else {
+						t.Fatalf("Test returned MountError %s, but expected error is not of MountError", got.Type)
+					}
 				} else if test.expectedError.Error() != publishErr.Error() {
 					t.Fatal(fmt.Errorf("expected error %q but got %q", test.expectedError.Error(), publishErr.Error()))
+					return
```
jooola, line 186: Might be unneeded since we have a Fatal just above.

test/integration/volumes_test.go
```diff
@@ -85,32 +101,50 @@ func TestVolumePublishUnpublish(t *testing.T) {
 
 				luksDevicePath := volumes.GenerateLUKSDevicePath(luksDeviceName)
 
-				return svc.FormatDisk(luksDevicePath, "ext4")
+				tmppath, err := os.MkdirTemp(os.TempDir(), "csi-driver-prepare")
+				if err != nil {
+					return err
+				}
```
jooola, line 107: Maybe you want to put this in a test helper, we use it quite often :)

jooola (approved):
Approved from my side :rocket:

apricote:
> Note that this change opens up the possibility to format the disk with almost any file system (mkfs.*), maybe the docs should be updated?

Looking at `k8s.io/mount-utils.SafeFormatAndMount.formatAndMountSensitive` it looks like it only supports `ext4`,`ext3` and `zfs`. Right now we also support `btrfs`, so this would break current functionality :(

test/integration/integration_test.go
```diff
@@ -105,3 +107,14 @@ func (w TestingWriter) Write(p []byte) (n int, err error) {
...
 }
+
+func formatDisk(mounter *mount.SafeFormatAndMount, device string, fstype string) error {
+	tmppath, err := os.MkdirTemp(os.TempDir(), "csi-driver-prepare")
+	if err != nil {
+		return err
+	}
+
+	defer os.RemoveAll(tmppath)
+	defer mounter.Unmount(tmppath)
```
apricote, line 118:
```suggestion
  // The library we use to format volumes only supports a combined "FormatAndMount"
  // so we call that and immediately unmount and cleanup the temp path afterwards.
	defer os.RemoveAll(tmppath)
	defer mounter.Unmount(tmppath)
```

test/integration/volumes_test.go
```diff
@@ -284,54 +299,56 @@ func TestDetectDiskFormat(t *testing.T) {
 
 	tests := []*struct {
```
apricote, line 300: Isnt `TestDetectDiskFormat` now unecessary because we only use the function exported from `k8s.io/mount-utils` or do you want to make sure that it still detects our LUKS disks correctly?

lukasmetzner: Yes, I kept the tests to ensure that our LUKS disks still work correctly.

PR #762, opened 2024-10-30
ci: fixed verify snapshots using wrong path

lukasmetzner:
In our verify snapshots step, the git diff command is targeting the deploy/ directory instead of chart/.snapshots/.


apricote (approved):

jooola:
This is a good candidate for a pre-commit hook :) if you want to add it to the csi-driver.

PR #775, opened 2024-11-11
refactor: redefintion and shadowing of builtin id cap

lukasmetzner:
Update PR of golangci-lint to v1.62.0 discovered a redefinition of cap in our capability check. Furthermore, we are shadowing cap in certain places.

Additionally, upgrade from v1.62.0 of golangci-lint (#774) can not be merged before this is fixed.


jooola:
You could have squeezed that one in the renovate upgrade PR, and merge it yourself, but that works too  :)

jooola (approved):

PR #780, opened 2024-11-12
feat: Added new option enableProvidedByTopology

lukasmetzner:
We are reintroducing a feature originally present in v2.10.0 to prevent pods from getting stuck in the `pending` state in clusters with non-cloud nodes. This feature is now optional and can be enabled via the Helm Chart. By default, it remains disabled to avoid compatibility issues with Nomad clusters, which have a different CSI spec implementation.

Learn more about it in #400.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -11,6 +11,9 @@ global:
   ##   - myRegistryKeySecretName
   ##
   imagePullSecrets: []
+  ## @param node.enableProvidedByTopology Enables workaround for upstream Kubernetes issue where nodes without the CSI plugin are still considered for scheduling.
+  ##
+  enableProvidedByTopology: false
 
 ## @section Common parameters
 ##
@@ -648,6 +651,7 @@ node:
   ## kubeletDir: /var/lib/k0s/kubelet
   ## For microk8s:
   ## kubeletDir: /var/snap/microk8s/common/var/lib/kubelet
+  ##
   kubeletDir: /var/lib/kubelet
 
 ## @section Other Parameters
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -12,6 +12,11 @@ global:
   ##
   imagePullSecrets: []
   ## @param node.enableProvidedByTopology Enables workaround for upstream Kubernetes issue where nodes without the CSI plugin are still considered for scheduling.
+  ## ref: https://github.com/kubernetes-csi/external-provisioner/issues/544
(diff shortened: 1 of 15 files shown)


docs/kubernetes/README.md
```diff
@@ -209,9 +209,17 @@ $ kubectl apply -f https://raw.githubusercontent.com/hetznercloud/csi-driver/v2.
 
 ## Integration with Root Servers
 
-Root servers can be part of the cluster, but the CSI plugin doesn't work there.
+Root servers can be part of the cluster, but the CSI plugin doesn't work there and the current behaviour of the scheduler can cause Pods to be stuck in `Pending`.
 
-Labels are needed to indicate whether a node is a cloud VM or a dedicated server from Robot. If you are using the `hcloud-cloud-controller-manager` version 1.21.0 or later, these labels are added automatically. Otherwise, you will need to label the nodes manually.
+To address this behavior, you can set `enableProvidedByTopology` to `true` in the Helm Chart configuration. This setting prevents pods from being scheduled on nodes — specifically, Robot servers — where Hetzner volumes are unavailable. Enabling this option adds the `instance.hetzner.cloud/provided-by` label to the [allowed topologies](https://kubernetes.io/docs/concepts/storage/storage-classes/#allowed-topologies) section of the storage classes that are created. Additionally, this label is included in the `topologyKeys` section of `csinode` objects, and a node affinity is set up for each persistent volume.
```
apricote, line 214: Please add a section that explains the backwards-incompatibility of this change and why users should not modify the setting in existing clusters.

chart/values.yaml
```diff
@@ -11,6 +11,9 @@ global:
   ##   - myRegistryKeySecretName
   ##
   imagePullSecrets: []
+  ## @param node.enableProvidedByTopology Enables workaround for upstream Kubernetes issue where nodes without the CSI plugin are still considered for scheduling.
```
apricote, line 14: - Can you link the upstream kubernetes issue here?
- Can you explain that this setting should not be modified for existing clusters (and why)

lukasmetzner:
@resmo I don't have much Nomad knowledge yet, but it might be enough to specify an anti-affinity towards the non-cloud nodes on the CSI Node Job. 

This should also suffice in Kubernetes, but there is an [upstream issue](https://github.com/kubernetes-csi/external-provisioner/issues/544) in the scheduler, which we built this workaround for.

apricote (approved):

PR #932, opened 2025-04-11
feat: allow to set labels for all volumes

cgroschupp:
This pull request adds support for setting default labels on all newly created volumes in the Hetzner Cloud CSI driver.

## Changes

- Introduced the environment variable `HCLOUD_VOLUME_DEFAULT_LABELS`, which accepts a comma-separated list of key:value pairs.
- Added the `ParseEnvMap` function to parse and validate the environment variable into a `map[string]string`.
- Extended `VolumeService` to apply the parsed labels during volume creation.
- Added unit tests for the `ParseEnvMap` function in `internal/app/app_test.go`.

## Notes

- Labels are only applied if `HCLOUD_VOLUME_DEFAULT_LABELS` is set and contains valid key:value pairs.
- Invalid formats in the environment variable will cause the controller to fail on startup.


cgroschupp:
I looked at some other CSI controller label/tag implementations and liked a combination of a static label from the controller via env/arg and the ability to specify the label via the storage class. From my side, at the moment I just need a way to specify only one label on a volume to identify which cluster it comes from. And I think it makes sense to define it as an arg/env variable for e.g. HCLOUD_CLUSTER_ID to set a clusterID label on all volumes.

I also like the feature to use "text/template" to allow the user to use variables within the label definition for e.g. the pvc name or namespace to the labels.

I'll see how time-consuming it is to implement this.

cgroschupp:
I have now added multiple ways to add labels to volumes. Each volume now gets `kubernetes.io/created-for/pvc/name`, `kubernetes.io/created-for/pvc/namespace`, `kubernetes.io/created-for/pv/name` and `csi.hetzner.cloud/created-by=csi-driver`. And it is possible to add labels via the storage class using the `labels` parameter and/or the `HCLOUD_VOLUME_EXTRA_LABELS` environment variable, for example for the clustername to be applied to all volumes.

My example helm values:
```
storageClasses:
  - name: hcloud-volumes
    defaultStorageClass: true
    reclaimPolicy: Delete
    extraParameters:
      labels: mylabel=myvalue
controller:
  volumeExtraLabels:
    cluster: cosmos
```

And a screenshot of the labels: 
image copy

lukasmetzner (changes requested):

cmd/aio/main.go
```diff
@@ -83,11 +84,18 @@ func main() {
 			hcloudClient,
 		),
 	)
+
+	extraVolumeLabels, err := utils.ConvertLabelsToMap(os.Getenv("HCLOUD_VOLUME_EXTRA_LABELS"))
+	if err != nil {
+		logger.Error("could not parse extra labels for volumes")
+		os.Exit(1)
+	}
```
apricote, line 92: Please make sure to log the error that happened during parsing. Otherwise, the user will not know how to fix the issue.

(Also applies to `cmd/controller/main.go`)

docs/kubernetes/README.md
```diff
@@ -155,6 +155,38 @@ During the initialization of the CSI controller, the default location for all vo
 3. If neither of the above is set, the `KUBE_NODE_NAME` environment variable defaults to the name of the node where the CSI controller is scheduled. This node name is then used to query the Hetzner API for a matching server and its location.
 4. As a final fallback, the [Hetzner metadata service](https://docs.hetzner.cloud/#server-metadata) is queried to obtain the server ID, which is then used to fetch the location from the Hetzner API.
 
+### Volume Labels
+
+It is possible to set labels for all newly created volumes. By default, all volumes are labeled as follows:
+
+* `kubernetes.io/created-for/pvc/name`
+* `kubernetes.io/created-for/pvc/namespace`
+* `kubernetes.io/created-for/pv/name`
```
apricote, line 164: These are not [well-known labels](https://kubernetes.io/docs/reference/labels-annotations-taints/) under `kubernetes.io`. I would prefer to use something else. Either the labels from the csi `external-provisioner` or custom ones under `csi.hetzner.cloud`.

- `csi.storage.k8s.io/pvc/name`
- `csi.storage.k8s.io/pvc/namespace`
- `csi.storage.k8s.io/pv/name`

Or

- `csi.hetzner.cloud/pvc-name`
- `csi.hetzner.cloud/pvc-namespace`
- `csi.hetzner.cloud/pv-name`

docs/kubernetes/README.md
```diff
@@ -155,6 +155,38 @@ During the initialization of the CSI controller, the default location for all vo
...
+* `csi.hetzner.cloud/created-by=csi-driver`
+
+To add extra labels to all created volumes set `HCLOUD_VOLUME_EXTRA_LABELS` in the format `key=value,...`.
+This is also configurable from the helm chart by the value `controller.volumeExtraLabels`, e.g:
```
apricote, line 168:
```suggestion
This is also configurable from the Helm chart by the value `controller.volumeExtraLabels`, e.g:
```

docs/kubernetes/README.md
```diff
@@ -155,6 +155,38 @@ During the initialization of the CSI controller, the default location for all vo
...
+```yaml
+controller:
+  volumeExtraLabels:
+    cluster: myCluster
+    env: prod
+```
+
+It is also possible to set only labels on specific volumes created by a storage class. To do this, you need to set `labels` in the format `key=value,...` as `extraParameters` inside the storage class.
+
+There is an example to set the `labels` for the storage class over the helm chart values:
```
apricote, line 179:
```suggestion
There is an example to set the `labels` for the storage class over the Helm chart values:
```

internal/driver/controller.go
```diff
@@ -66,12 +84,38 @@ func (s *ControllerService) CreateVolume(ctx context.Context, req *proto.CreateV
 		location = *loc
 	}
 
+	var volumeLabels = map[string]string{
+		tagKeyCreatedBy: "csi-driver",
```
apricote, line 88: Not sure if `csi-driver` gives any additional information here. The label `csi.hetzner.cloud/created-by` is already unique. As labels can have an empty value, I would suggest using that instead.

@lukasmetzner What do you think?

internal/driver/controller.go
```diff
@@ -66,12 +84,38 @@ func (s *ControllerService) CreateVolume(ctx context.Context, req *proto.CreateV
...
+			volumeLabels[tagKeyCreatedForVolumeName] = value
+		case parameterKeyLabels:
+			customLabels, err := utils.ConvertLabelsToMap(value)
+			if err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "Invalid format of parameter labels: %s", err)
+			}
+			maps.Copy(volumeLabels, customLabels)
+		default:
+			return nil, status.Errorf(codes.InvalidArgument, "Invalid parameter key %s for CreateVolume", key)
+		}
```
apricote, line 109: We previously did not validate that the parameters are from a known list, so adding this now would be a breaking change. Please ignore unknown parameters instead.

internal/utils/labels.go
```diff
@@ -0,0 +1,31 @@
...
+	result := map[string]string{}
+	splitFn := func(c rune) bool {
+		return c == ','
+	}
+	vals := strings.FieldsFunc(labelsString, splitFn)
+	for _, val := range vals {
+		pair := strings.SplitN(val, "=", 2)
+		if len(pair) < 2 {
+			return nil, fmt.Errorf("invalid value %s", val)
+		}
```
apricote, line 19: Labels may have an empty value, so this should be handled as if the value was an empty string instead of an error.

internal/utils/labels.go
```diff
@@ -0,0 +1,31 @@
...
+		key := strings.TrimSpace(pair[0])
+		if key == "" {
+			return nil, errors.New("empty key")
+		}
+		value := strings.TrimSpace(pair[1])
+		if value == "" {
+			return nil, errors.New("empty value")
+		}
```
apricote, line 27: As mentioned above, label value may be empty.

apricote (changes requested):
:rocket: Very nice in general, thanks for adding tests!

cgroschupp:
@apricote The pipeline is not running due to the following reasons: `/home/runner/work/_actions/hetznercloud/tps-action/main/get-token.sh: line 27: ACTIONS_ID_TOKEN_REQUEST_URL: unbound variable`

cgroschupp:
> Can you take a look at the issues from the [lint](https://github.com/hetznercloud/csi-driver/actions/runs/14852822901/job/41707564290?pr=932) and [pre-commit](https://github.com/hetznercloud/csi-driver/actions/runs/14852822901/job/41707564316?pr=932) jobs? 

@apricote done

internal/driver/controller.go
lukasmetzner: I can see this being helpful if a user wants to perform certain actions (protection, deletion) on all csi-driver managed volumes.

PR #955, opened 2025-05-07
ci(e2e): less ginkgo output

apricote:
The `e2e.test` binary has __all__ kubernetes tests, but we only run those that start with `External.Storage`. This means there is a lot of "Skipped" output right now, that is useless. By setting `--silence-skips` these log lines are removed.

Additional ginkgo offers a `--github-output` flag that promises to make the output nicer for GitHub actions. Not sure what it really does.

These flags are only supported in Ginkgo 2.18+, which is only available in Kubernetes 1.31+.


lukasmetzner (approved):

apricote:
This reduced the log lines for the parallel tests from ~6700 to ~4900.

PR #988, opened 2025-06-03
feat: rename default labels on Volumes

lukasmetzner:
After implementing the new label we have discovered an inconsistency in the Hetzner Cloud public API. The topic was discussed internally and as a result we have to remove the `csi.hetzner.cloud` prefix from the label. Besides, we have decided to rename the label to `managed-by`, as this is more consistent with other integrations (e.g., fleeting-plugin-hetzner). 

Additionally, the Kubernetes labels (e.g., `csi.storage.k8s.io/pvc/name`) are not valid, as there is a second `slash` in the key.

### DIFF
--- a/internal/driver/controller.go
+++ b/internal/driver/controller.go
@@ -24,9 +24,9 @@ const (
 	parameterKeyPVName       = "csi.storage.k8s.io/pv/name"
 	parameterKeyLabels       = "labels"
 
-	labelKeyPVCName      = "pvc-name"
-	labelKeyPVCNamespace = "pvc-namespace"
-	labelKeyPVName       = "pv-name"
+	labelKeyPVCName      = "csi.storage.k8s.io/pvc-name"
+	labelKeyPVCNamespace = "csi.storage.k8s.io/pvc-namespace"
+	labelKeyPVName       = "csi.storage.k8s.io/pv-name"
 	labelKeyManagedBy    = "managed-by"
 )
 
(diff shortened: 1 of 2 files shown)


internal/driver/controller.go
```diff
@@ -24,10 +24,10 @@ const (
...
 	parameterKeyLabels       = "labels"
 
-	tagKeyCreatedForClaimName      = "csi.storage.k8s.io/pvc/name"
-	tagKeyCreatedForClaimNamespace = "csi.storage.k8s.io/pvc/namespace"
-	tagKeyCreatedForVolumeName     = "csi.storage.k8s.io/pv/name"
-	tagKeyCreatedBy                = "csi.hetzner.cloud/created-by"
+	labelKeyPVCName      = "csi.storage.k8s.io/pvc-name"
+	labelKeyPVCNamespace = "csi.storage.k8s.io/pvc-namespace"
+	labelKeyPVName       = "csi.storage.k8s.io/pv-name"
+	labelKeyManagedBy    = "managed-by"
```
lukasmetzner, line 30: I have updated the labels, but I find this to be a bit inconsistent (prefix + no prefix), and we also change keys of a domain we do not own (`pvc/name -> pvc-name`.

What are your thoughts @apricote? 

Otherwise, we could also go for:
```
pvc-name
pvc-namespace
pv-name
managed-by
```

lukasmetzner: Decision we remove the k8s.io prefixes.

PR #996, opened 2025-06-06
feat: single binary

lukasmetzner:


deploy/kubernetes/hcloud-csi.yml
```diff
@@ -214,7 +214,9 @@ spec:
         - name: hcloud-csi-driver
           image: docker.io/hetznercloud/hcloud-csi-driver:v2.15.0 # x-releaser-pleaser-version
           imagePullPolicy: IfNotPresent
-          command: [/bin/hcloud-csi-driver-node]
+          command: [/bin/hcloud-csi-driver]
+          args:
+            - -node
```
apricote, line 219: Changing the command and required `args` is kind of a breaking change for users that are just bumping the image tag but are not pulling the manifests for each release.

I do not think we have a fixed policy for this, but at a minimum it needs to be mentioned in the release notes.

---

When we are changing this anyway, should we just add an `ENTRYPOINT /bin/hcloud-csi-driver` and drop the `command` from the manifest?

apricote (commented):
The new binary is the same size as one of the previous binaries (15MB), so this drops the combined docker image by 15MB :rocket:

apricote (approved):
:package:

Issue #1149, opened 2025-11-07
bug(regression): merging controller and csi node broke app in ipv6 only setups


hegerdes:
### TL;DR

App can not fetch required info in ipv6 only due to reliance on metadata service - caused by #996 

### Expected behavior

I can run the controller and csi-node on ipv6 only environments.  
Required information should be tried from multiple sources.

### Observed behavior

Controller and node crash on ipv6 only pods due to missing info about location and server-id

### Minimal working example

Create a single node cluster with ipv6 on hcloud

### Log output

```plain text
Logs:

time=2025-11-07T21:28:13.441Z level=DEBUG source=/home/runner/work/csi-driver/csi-driver/internal/metrics/metrics.go:36 msg="registering metrics with registry"
time=2025-11-07T21:28:13.441Z level=DEBUG source=/home/runner/work/csi-driver/csi-driver/internal/metrics/metrics.go:43 msg="registered metrics"
time=2025-11-07T21:28:13.441Z level=DEBUG source=/home/runner/work/csi-driver/csi-driver/internal/metrics/metrics.go:69 msg="starting prometheus http server" addr=0.0.0.0:9189
time=2025-11-07T21:28:13.442Z level=WARN source=/home/runner/work/csi-driver/csi-driver/cmd/main.go:51 msg="unable to connect to the metadata service"
time=2025-11-07T21:28:13.442Z level=ERROR source=/home/runner/work/csi-driver/csi-driver/cmd/main.go:56 msg="could not determine default volume location" error="HCLOUD_VOLUME_DEFAULT_LOCATION not set and not running on a cloud server"
stream closed: EOF for kube-system/hcloud-csi-controller-55d6888b97-mn2fh (hcloud-csi-driver)
```

### Additional information

Previously the app used multiple sources to get the required information about location and server ID. Now it only relies on hardcoded information or the (ipv4 only) metadata-service.

The app should try to fetch all needed information with fallbacks in place. The suggested order:
 * Explicit configuration via ENV or Conf-File
 * Query the metadata-service
 * Query the Hetzner-API
 * Query the Kubernetes Node-API (via labels set by the ccm)

The first three steps were already used before merging the node and controller. In the file https://github.com/hetznercloud/csi-driver/blame/main/internal/app/app.go is still the old code to query by Hetzner-API as a fallback

hegerdes:
Might be a duplicate of #95 but I think I got it working before. Nevertheless would be nice to have fallbacks

PR #1158, opened 2025-11-12
fix: missing location config via node name and server id

lukasmetzner:
When merging our two binaries into one, the option to configure
the location on startup via KUBE_NODE_NAME or HCLOUD_SERVER_ID was not implemented.

Missing configuration options in docs: https://github.com/hetznercloud/csi-driver/blob/637e45cf283f384c455125bbc7f3931328f5e80b/docs/kubernetes/README.md?plain=1#L154-L155

Fixes https://github.com/hetznercloud/csi-driver/issues/1149

### DIFF
--- a/test/integration/location_test.go
+++ b/test/integration/location_test.go
@@ -1,3 +1,5 @@
+//go:build integration
+
 package integration
 
 import (


lukasmetzner:
> Is there an easy way to test this code?

I could think of an integration test where we spin up a test server and a fake metadata server, to test all four paths of `GetServerLocation`.

PR #1180, opened 2025-11-24
ci: use makefile to template snapshots and deploy-manifests

lukasmetzner:


lukasmetzner:
Includes some changes to snapshots and deploy-manifests. Both were created with `--kube-version v1.31.4 `, which is EOL.

PR #1183, opened 2025-11-24
chore: drop k8s v1.31 support

lukasmetzner:
```rp-commits
fix: drop Kubernetes v1.31 support
```


apricote (approved):

lukasmetzner:
> Why `chore`, shouldn't we communicate this the users?

Was EOL a month ago, so we usually drop it when adding the new version. We forgot that in this case. But you are right, we should mention it in the next release note. I will rewrite it.

PR #1360, opened 2026-05-12
fix(chart): prefix local templates to avoid bitnami common collision

casinesque:
Opening this PR to fix a named template collision issue that may affect users who include this chart as a dependency in umbrella charts alongside bitnami/common v2.

**Problem**: The chart vendors a local copy of bitnami/common helper templates inside the chart/templates/ directory. These templates share the same signature as the official bitnami/common library. When both are present in an umbrella chart, Helm has no collision resolution mechanism and picks the local v1 template over the v2 one, breaking rendering since the two versions expect different arguments.

**Fix**: Renamed all internal named templates with the hetzner prefix. This makes the two sets of templates unambiguous to Helm, allowing them to coexist without conflicts.

**Backwards compatibility**: This fix is safe regardless of which version of bitnami/common is used alongside this chart. Since the renamed templates no longer share a signature with either v1 or v2, both can now coexist in the same umbrella chart without interference.

**Testing**: This PR only affects rendered chart output. The Go tests fail on main as well (unrelated to this change), so they are out of scope here. Validation was done by diffing helm template output between main and this branch  and the rendered manifests are identical, confirming the renamed templates resolve correctly.


lukasmetzner:
@casinesque Thanks for the contribution. Could we please use the prefix `hcloud-csi.*` instead of `hetzner.common.*`.

Could you please also remove the trailing newline from the `.gitignore` file. You can check the pre-commit CI job for that.

For your information, the e2e tests are expected to fail here, due to restrictive permissions in the CI.

casinesque:
@lukasmetzner thanks for your consideration and tips.
 
I synced my fork and branch with upstream, i made the `hcloud-csi.*` replacements and `.gitignore` should be fixed now. 

Let me know if anything else needs to be improved, thanks!

lukasmetzner (approved):

PR #1364, opened 2026-05-13
fix: recover stale luks device mapper

lukasmetzner:


lukasmetzner:
Debugging information provided by customer indicates there is an issue with luks device mappers not being closed properly. I could not yet reproduce this correctly, but this would address the issue.

lukasmetzner:
As an addition, I properly introduced context management into the cryptsetup. Required a little refactor, but this is mostly just dragging along the context.

Jank1310:
@lukasmetzner I think we got hit by somthing related to this pr
Happy to share the full `hcloud-csi-node` logs and `cryptsetup status` if needed

AI
----

A dm-crypt mapper left over from an earlier attach cycle stayed active while pointing at a dead `dev_t` — `cryptsetup status` reported `type: n/a` / `device: (null)`. Since `Open()` short-circuits on `IsActive()`, which only inspects cryptsetup's exit code and discards its output, the driver reused that zombie mapper; mount-utils then concluded it was unformatted and ran `mkfs.ext4 -F` against it — **708 times over 23.8 hours**, each attempt failing with I/O errors and emitting a fresh `Filesystem UUID`.

I ran this PR's `NewLUKSDeviceStatus`/`IsZombie()` against our verbatim `cryptsetup status` output and it returns **true**, so #1364 would have closed the stale mapper and re-opened LUKS against the real device instead of mkfs-ing a dead one.

PR #1381, opened 2026-05-29
fix: device needs to be ready on node publish

lukasmetzner:


jooola (approved):

internal/volumes/mount.go
```diff
@@ -153,6 +163,43 @@ func (s *LinuxMountService) Publish(ctx context.Context, targetPath string, devi
...
 
+// waitDeviceReady ensures the device at devicePath exists. This is done by ensuring a stat
+// syscall returns no error.
+func (s *LinuxMountService) waitDeviceReady(ctx context.Context, devicePath string) error {
+	const maxRetries = 7
+	backoffFunc := hcloud.ExponentialBackoffWithOpts(hcloud.ExponentialBackoffOpts{
+		Base:       time.Millisecond * 50,
+		Multiplier: 2.0,
+		Cap:        500 * time.Millisecond,
+	})
```
jooola, line 174: With a 500ms cap, we might as well just wait 500ms every time. What is the reason for having such a small cap? Could we increase the cap to e.g. 5s?

(Without cap, we would have a total waiting time of 12.8 seconds, with a max of 7 retries)

lukasmetzner: We don't fully understand yet why this device appears late. Talking with internal teams, one idea is that the OS is under heavy pressure. This seems to happen very rarely, as we don't have a lot of issue reports from customers.

My idea was to quickly retry in small intervals, to absorb the delays caused by the OS being busy.

Issue #1455, opened 2026-08-10
NodePublishVolume can mount the wrong volume: waitDeviceReady only checks that the by-id path exists


victorlane:
## What happened

After a node reboot, two PVCs were published onto the same block device. `docmost/data-docmost-redis-0` and `monitoring/monitoring-grafana` both ended up on the device holding the Grafana volume. Redis then wrote its AOF into the Grafana volume, and because the container runtime relabelled the shared device for one pod, the other pod lost access with `Permission denied` (SELinux MCS). Both workloads stayed broken for 2 days until the pods were deleted, which caused a correct republish.

## Environment

- csi-driver v2.21.2
- RKE2 v1.36.3, openSUSE Leap Micro 6.2, SELinux enforcing
- Volumes already attached to the node at boot

## Evidence

Node plugin log for that boot. Note the readiness retries, then two successful publishes 112 ms apart:

```
00:03:14 ERROR failed to publish volume: device "/dev/disk/by-id/scsi-0HC_Volume_106486781" not ready: no such file or directory
00:03:14 ERROR failed to publish volume: device "/dev/disk/by-id/scsi-0HC_Volume_106478890" not ready: no such file or directory
(further retries)
00:03:27.670 INFO publishing volume target-path=.../pvc-dc7df2e4-.../mount device-path=/dev/disk/by-id/scsi-0HC_Volume_106486781
00:03:27.782 INFO publishing volume target-path=.../pvc-742f8b5a-.../mount device-path=/dev/disk/by-id/scsi-0HC_Volume_106478890
```

The resulting mounts point at the same device (8:32), even though the recorded source paths differ:

```
# findmnt -rno SOURCE,MAJ:MIN,TARGET | grep kubernetes.io~csi
/dev/disk/by-id/scsi-0HC_Volume_106486781 8:32 /var/lib/kubelet/pods/.../pvc-dc7df2e4-.../mount
/dev/disk/by-id/scsi-0HC_Volume_106478890 8:32 /var/lib/kubelet/pods/.../pvc-742f8b5a-.../mount
```

While the symlinks resolve like this:

```
/dev/disk/by-id/scsi-0HC_Volume_106478890 -> /dev/sdc   (8:32)
/dev/disk/by-id/scsi-0HC_Volume_106486781 -> /dev/sdd   (8:48)
```

[...]

Issue #1346, opened 2026-04-30
Volume data destroyed — blkid incorrectly reports formatted ext4 volume as unformatted after node migration, triggering mkfs


Manuelraa:
### TL;DR

Pod was reschedules from node1 to node2. `hccm-hcloud-csi-node` wrongly reformat disk

### Expected behavior

Disk is not reformatted.

### Observed behavior

Disk gets reformatted even though it was already in use for 318 days.

### Minimal working example

Can't reproduce. Worked for a year without issues across many PVCs.

### Log output

```plain text
# CSI node plugin on node1:

17:54:16 mount.go:161  unpublishing volume  device-path=/dev/disk/by-id/scsi-0HC_Volume_<volume-id>

# Volume detached from node1, attached to node2:

17:54:43 mount.go:125  publishing volume    device-path=/dev/disk/by-id/scsi-0HC_Volume_<volume-id>  fs-type=ext4
# ^^^ succeeds normally on node2 — no format triggered

# ~16 minutes later, pod evicted again, volume moves back to node1:

18:10:56 mount.go:161  unpublishing volume  [on node2]
18:11:25 mount.go:125  publishing volume    device-path=/dev/disk/by-id/scsi-0HC_Volume_<volume-id>  fs-type=ext4  [on node1]
18:11:25 mount_linux.go:608  Disk "/dev/disk/by-id/scsi-0HC_Volume_<volume-id>" appears to be unformatted, attempting to format as type: "ext4" with options: [-F -m0 /dev/disk/by-id/scsi-0HC_Volume_<volume-id>]
18:11:25 mount_linux.go:619  Disk successfully formatted (mkfs): ext4 - /dev/disk/by-id/scsi-0HC_Volume_<volume-id>
```

### Additional information

More verbose logs showing successful attempt and then the catastrophic failure.
[volume-102736836-event-chain.log](https://github.com/user-attachments/files/27244382/volume-102736836-event-chain.log)

PV spec (relevant fields)
```
spec:
  csi:
    driver: csi.hetzner.cloud
    volumeHandle: "<volume-id>"
    fsType: ext4
    volumeAttributes:
      fsFormatOptions: ""
  storageClassName: hcloud-volumes
  accessModes: [ReadWriteOnce]
  capacity:
    storage: 10Gi
```

Jank1310:
[...]

PR #1456, opened 2026-08-10
fix: verify device identity before publishing volume

victorlane:
## Problem

`NodePublishVolume` trusts the device path from `PublishContext` verbatim. `waitDeviceReady` only checks that the `/dev/disk/by-id` path exists, not that it points at the requested volume. The symlinks are maintained asynchronously by udev, so at boot with several volumes attaching, a path can resolve to another volume's block device.

Publish then proceeds on the wrong device:

- If it carries a filesystem, it is mounted and two volumes share one disk (hetznercloud/csi-driver#1455, production incident with SELinux lockout and cross-volume writes).
- If blkid finds no filesystem, it is formatted. The data is destroyed even when the subsequent mount fails. Same outcome as #1346, reached through a different route.

Both were reproduced against current `main` at the `Publish` level (mkfs ran and zeroed a data-carrying aliased device).

## Fix

After the existing stat check, resolve the symlink and compare the device's SCSI serial (`/sys/block/<dev>/device/vpd_pg80`, which Hetzner sets to the volume ID) against the ID in the path. A mismatch is retried inside the existing backoff loop, treating a stale symlink like one that does not exist yet, and ends in `ErrDeviceMismatch` instead of a wrong mount.

Verification fails open: non Hetzner paths, LUKS mappings, or an unreadable VPD page skip the check, so a mount that cannot be verified is never blocked. The VPD layout (4 byte header, then the volume ID) was confirmed against production nodes.

## Tests

- `TestPublishRejectsAliasedDevice`: aliased by-id path at the `Publish` entry point, asserts `ErrDeviceMismatch` and that the aliased device's data is untouched. Needs a writable `/dev`, skips otherwise.
- `TestVerifyDeviceSerial`: match, mismatch, kernel padded serial, unverifiable device.
- `TestVerifyDeviceIdentitySkipsForeignPaths`: LUKS mappings and the sanity tests' fake `devpath` are never rejected.

Full suite passes; `go vet` and `gofmt` clean.

Refs hetznercloud/csi-driver#1455

This PR (and code) were written using Claude Opus 5 assistance

### DIFF
--- a/internal/volumes/mount.go
+++ b/internal/volumes/mount.go
@@ -1,6 +1,7 @@
 package volumes
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


lukasmetzner:
Hey, before going any further, please have a quick look at our contributing guideline, in particular the AI/LLM policy: https://github.com/hetznercloud/csi-driver?tab=contributing-ov-file#ai--llm-policy

I suspect this PR was created with the assistance of an AI tool, but this was not disclosed.

lukasmetzner:
`Test / kubernetes * (pull_request)` `Test / nomad * (pull_request)` failing is expected.

victorlane:
> Hey, before going any further, please have a quick look at our contributing guideline, in particular the AI/LLM policy: https://github.com/hetznercloud/csi-driver?tab=contributing-ov-file#ai--llm-policy
> 
> I suspect this PR was created with the assistance of an AI tool, but this was not disclosed.

Added it

PR #1457, opened 2026-08-12
fix: verify device identity before mounting

### LINKED ISSUE #1346: Volume data destroyed — blkid incorrectly reports formatted ext4 volume as unformatted after node migration, triggering mkfs

lukasmetzner:
The `/dev/disk/by-id` symlinks are maintained asynchronously by udev, so
after an attach that reuses a device name, the symlink of one volume can
still resolve to the block device of another.

The resolved device is now checked against the volume ID taken from the
by-id path by reading its SCSI serial from VPD page 0x80 in
`/sys/class/block/<device>/device/vpd_pg80`. A mismatch is retried
through the existing `waitDeviceReady` backoff, which is where a symlink
udev has not caught up with yet becomes correct.

The check is fail-closed: if the serial cannot be read, the volume is
not published rather than published unverified, because mounting the
wrong device destroys data while refusing to mount does not. Devices
that do not expose `vpd_pg80` are therefore no longer published.

Fixes #1346

### DIFF
--- a/internal/driver/node.go
+++ b/internal/driver/node.go
@@ -62,8 +62,12 @@ func (s *NodeService) NodePublishVolume(ctx context.Context, req *proto.NodePubl
 		return nil, status.Error(codes.InvalidArgument, "missing target path")
 	}
 
-	devicePath := req.GetPublishContext()["devicePath"]
-	if devicePath == "" {
+	// The device path the controller published is not used to find the device: it names
+	// a by-id symlink, which udev maintains asynchronously, so it can still point at the
+	// device of another volume. The device is looked up by the serial the kernel reports
+	// for it instead. The publish context entry is still required, as its presence is how
+	// the CO signals that the volume was attached to this node.
+	if req.GetPublishContext()["devicePath"] == "" {
 		return nil, status.Error(codes.InvalidArgument, "missing device path")
 	}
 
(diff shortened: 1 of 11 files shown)


victorlane:
@lukasmetzner what is the status on this fix? We've ran into this issue twice again.

lukasmetzner:
@victorlane Hey, apologies we are currently low on capacity in the team and I lost sight of the issue. There are couple of things I would like to polish, but then this should be good to go. Currently on vacation, will pick this up as soon as I am back :)

PR #1498, opened 2026-09-16
feat: propose native volume snapshot support

### LINKED ISSUE #849: VolumeSnapshot support for Backup solutions
### TL;DR

Nowadays most csi-drivers are supporting VolumeSnapeshots as of this list: https://kubernetes-csi.github.io/docs/drivers.html
In order to have an independent backup solution of kubernetes clusters running on hetzner this support would be required.

### Expected behavior

The csi-driver should be capable of creating VolumeSnapshots an should be used in VolumeSnapshotClasses.

eugene-harold-krabs:
## Proposal

Implement native CSI block-volume snapshots on top of the companion Cloud API
client proposal in hetznercloud/hcloud-go#921.

This is intentionally a draft. The driver compiles against that exact proposal
commit through a temporary module replacement; the replacement and its narrow
lint allowance must be removed once `hcloud-go` publishes the accepted API.

## Driver behavior

- Advertises `CREATE_DELETE_SNAPSHOT` and `LIST_SNAPSHOTS`.
- Implements idempotent `CreateSnapshot` and `DeleteSnapshot` plus filtered,
  paginated `ListSnapshots`.
- Restores `CreateVolume` from a snapshot, validates capacity and topology, and
  retains the CSI content source in the response.
- Records the source snapshot ID on restored volumes so repeated requests cannot
  silently reuse a same-named volume created from different content.
- Waits for provider actions and reports a snapshot ready only after the native
  resource reaches `available`.
- Keeps snapshots crash-consistent; application quiescing stays with the caller.

## Kubernetes integration

- Adds the official `csi-snapshotter` v8.6.0 sidecar to the controller chart.
- Adds the required snapshot RBAC, configurable image/resources/arguments,
  schema entries, rendered chart snapshots, and deployment manifest.
- Documents `VolumeSnapshotClass`, snapshot creation, restore, capacity, and the
  cluster-wide CRD/snapshot-controller prerequisite.

## Validation

- `go test ./...`
- `go vet ./...`
- `golangci-lint v2.13.2` — 0 issues
- `helm lint chart`
- regenerated Helm snapshots and deploy manifest

## Tracking

- Closes hetznercloud/csi-driver#849 once the provider API is shipped and this
  implementation is accepted.
- https://gitlab.com/xrow-public/infrastructure/-/work_items/46

Dependency: hetznercloud/hcloud-go#921


jooola:
Closing, see https://github.com/hetznercloud/csi-driver/issues/849#issuecomment-5694717485.

